PR #7, opened 2016-01-18
API-based approach, cache Environment in memory

paf31:
cc @garyb @sharkdp @kRITZCREEK

The plan is to host this on `compile.purescript.org`, and have `try.purescript.org` be hosted on GH Pages and call this API using CORS.

Caching the `Environment` makes the builds effectively instantaneous, which should allow us to build a really nice interface for beginners. I hope to integrate some `psc-ide` functionality in the UI eventually too. 

Thanks.


kRITZCREEK:
I can take a look at getting autocompletions from psc-ide. It would probably best be run as it's own process on the server so it doesn't crash the website in case of failure... :D I might need to add a command to load an extern file by filepath since the externs don't follow the "usual" `output-dir/Control.Monad/externs.json` file structure.
Looks like a cool way to introduce people to the language!

paf31:
I've added a basic client application which I still need to improve.

@kRITZCREEK If you could work on that, that would be amazing! My wishlist is:

- Types and docs for names under the cursor, from externs files
- Links to things like Pursuit (and the book for keywords like `data`)

sharkdp:
I just tested this quickly (with the -playground client). It builds and works well. The speed is absolutely amazing. Will play more with this in the next days..

paf31:
I'll deploy this later today, so people can try it out, and we can add the other features over time.

paf31:
[Here it is](http://community.purescript.org/try/). Let me know what you think.

sharkdp:
Is it somehow missing `Data.Prim`? I'm getting things like
```
  Unknown type Int
```

sharkdp:
Okay, I get things running if I

- add a (fake) externs file for `Prim`: https://gist.github.com/sharkdp/f35a38e0ebc767e35c56
- Add `"Prim"` in the `externsFiles` list in `Main.hs`

Unfortunately, I still have to `import Prim` in the client.

paf31:
Oops, I forgot to add the default `Prim` import. I'll do that today and redeploy.

If we can make a list of easily-fixed issues like this one, I'll aim to make a proper release this week.

I hope we can use the same API to make the examples on www.purescript.org interactive as well, once I've added a couple of extra libraries (quickcheck, smoulder and affjax jump to mind)

paf31:
- Mobile layout should be vertical.
- Stop flexbox columns from growing sideways.

sharkdp:
Not sure if it's easy to fix, but something is off with encoding/decoding. If you compile & run this:
``` purs
module Main where

import Prelude
import Control.Monad.Eff.Console (log)

main = log "°"
```

It prints `Â°`

paf31:
Ah, thanks for the report!

sharkdp:
> Let me know what you think.

Forgot to say: This is incredibly cool! :smile: 

After some struggling, I managed to get it running with the full `Flare` module (which also pulls in `DOM`, `Signal` and a few other additional modules). The `bundle.js` is now around `450kb` but it is still very fast (feels instantaneous).

paf31:
Great! Glad to hear things are still zippy after adding more modules. I'll put some instructions together soon for adding more modules, since I'd like people to be able to make PRs easily.

paf31:
I should probably add a list of available modules to the page somewhere too.

sharkdp:
> I'll put some instructions together soon for adding more modules, since I'd like people to be able to make PRs easily.

That would be great. It took me some time until I managed to find the right `psc` command to build the `bundle.js` and to manually do the topological sort inside `Main.hs` :joy:

paf31:
@sharkdp I've added Flare now, as well as Affjax. I just need to figure out a way to change out the HTML based on the example.

sharkdp:
> @sharkdp I've added Flare now, as well as Affjax.

Wow, Thanks!

> I just need to figure out a way to change out the HTML based on the example.

This probably requires a lot of work, but my favorite solution would be to allow several source files (like [JS Bin](https://jsbin.com/?html,js,output) or similar sites). In this way, one could add an editor tab for the HTML input. This would also allow to showcase the JavaScript FFI, by (simply) including a `Main.js` tab.

paf31:
If someone could help me to understand why the iframe is growing on iOS, but working in Chrome, I'd really appreciate it.

The main thing now seems to be to set up the initial page. I'm thinking I'll use Gist to host the other examples, and that the default code will use virtual DOM to render a list of links to other gists. Any thoughts on how we should wrap virtual DOM?

paf31:
The latest version is up, could I please get some feedback. The remaining issue I know of is a scrolling issue.

I also need to create more examples.

kRITZCREEK:
I think I hit the scrolling issue. I can't scroll down in the code editor. Looks like it doesnt "recalculate" the scrolling area when the rendering result comes back from the server and takes up space.

paf31:
Ok, the scroll issue is fixed now I think.

If anyone would like to contribute more examples, I'd be very grateful. I'm going to try to push this out at `try.purescript.org` this weekend.

paf31:
@sharkdp I've been thinking about packages to include again, and I've got another idea. Since it's so simple to spin up servers now, it might make sense to deploy other bundles separately. I'm thinking:

- Try PureScript (core libraries only), with links to:
- Try Flare (flare + signal + drawing)
- Try Thermite
- Try Neon
- Try Halogen

Thoughts? I'll deploy Try PureScript this weekend either way, but I can quickly deploy new servers if we decide this is a good approach. Clients can simply be forks of the UI portion.

paf31:
Here's a demo, which supports all of the core libraries, but no drawing etc.

paf31:
Could someone please review the code?

garyb:
It's a little tricky to do so: "Sorry, we could not display the entire diff because it was too big."

kRITZCREEK:
I noticed some `error` or `undefined` to supply optional arguments to compiler functions. Might be a motivation to make the purescript codebase a bit more modular?

paf31:
@kRITZCREEK True, but that's just the filename. I'm not too concerned about using `undefined` there right now, but it probably should be fixed at some point.

paf31:
@sharkdp The first release (core libraries) is live now, and I'm happy to work with you on deploying Try Flare if you like. I put some instructions for creating new configurations in the README.

sharkdp:
> Since it's so simple to spin up servers now, it might make sense to deploy other bundles separately.

Yes, I think that's a good idea.

> The first release (core libraries) is live now, and I'm happy to work with you on deploying Try Flare if you like. I put some instructions for creating new configurations in the README.

I'd be very interested in helping with this. I have a few questions regarding this:

- Would you like to host these alternative backends on the purescript server? That would be great. Otherwise, I could probably also host it on my VPS.
- I have trouble testing this locally with the current version, because you removed the `Access-Control-Allow-Origin` header from the server code at some point. How do you get around this when running the server?

paf31:
> Would you like to host these alternative backends on the purescript server?

Definitely. I'm working on Try Thermite right now, but I'll look at Flare after that, and if you can put the config file and externs together somewhere, I can deploy it for you.

I set the `Access-Control-Allow-Origin` header in Nginx.

paf31:
Try Thermite is up now, so I can spend time deploying Try Flare when you're ready.

PR #11, opened 2016-01-25
Use two spaces for tabs in the editor.

sharkdp:
Just a suggestion. You used two spaces in all the examples, as far as I can see.


paf31:
:+1:

PR #15, opened 2016-03-13
Bug fixes - no output in Firefox/Safari & crazy scroll bars

LiamGoodacre:
It appears that `innerText` is not supported in these browsers.
However `appendChild` and `createTextNode` appear to have better support: http://caniuse.com/#search=createTextNode

### DIFF
--- a/css/index.css
+++ b/css/index.css
@@ -4,16 +4,13 @@ html, body {
   height: 100%;
 }
 
-body
-{
+body {
   font-family: 'Roboto', sans-serif;
-  color: #404040;
   margin: 0;
   color: rgb(29, 34, 45);
 }
 
 #wrapper {
-  margin: 0 auto;
   display: flex;
   flex-direction: column;
   min-height: 100vh;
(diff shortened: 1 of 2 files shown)


LiamGoodacre:
Just noticed that it used to use `createTextNode` what was the reason for switching to `innerText`?
https://github.com/purescript/trypurescript/commit/27e34bcd69752bbb856085ecc0af3e0b8764f338#diff-42f69825340f0b5fa628d80900a46daaL63

LiamGoodacre:
This look okay @paf31 ? :smiley:

paf31:
Looks good, but I don't remember why I changed this. Did you test it in Chrome too?

LiamGoodacre:
Yes sorry I should have said: tested with Chrome, Firefox, and Safari.  But I don't have access to IE here.

LiamGoodacre:
Tried IE 11, but neither my updates nor the existing site display anything - so I haven't made it any worse there...

paf31:
Thanks! We can try and figure out IE11 separately then.

paf31:
I'm not sure if this works on iOS. The scrolling seems messed up.

Issue #14, opened 2016-02-04
Button to toggle auto compile


texastoland:
While a phenomenal feature, I wrote some laggy synchronous code that intermittently froze the editor when I typed too slow. Alternatively a longer delay might help e.g. half a second.

PR #23, opened 2016-06-24
Adding options to generate JS and view modes

soupi:
New version can be seen [here](https://soupi.github.io/trypurescript).

### DIFF
--- a/css/index.css
+++ b/css/index.css
@@ -18,8 +18,30 @@ body {
 
 #body {
   display: flex;
+  flex-direction: col;
+  flex: 1;
+}
+#options {
+  position: absolute;
+  top: 2px;
+  left: 2px;
+  right: 0;
+  bottom: 0;
+}
+#options input {
+  vertical-align: top;
+}
+
+#editor {
+  display: flex;
   flex-direction: row;
   flex: 1;
+  position: absolute;
+  top: 25px;
+  left: 0;
+  right: 0;
+  bottom: 0;
+
 }
 
 #column1, #column2 {
@@ -33,18 +55,9 @@ body {
   overflow: auto;
 }
 
-#options {
-  position: absolute;
-  top: 2px;
-  left: 1;
-  right: 2px;
-  bottom: 0;
- 
-}
-
 #code {
   position: absolute;
-  top: 25px;
+  top: 0;
   left: 0;
   right: 0;
   bottom: 0;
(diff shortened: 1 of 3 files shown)


paf31:
Thanks! A few observations after trying this:

- I would move the radio buttons to the right and show them as tabs.
- The "show js" option is only applicable when the output is shown, so it should be disabled otherwise.

js/index.js
```diff
@@ -12,6 +12,18 @@
...
 $(function() {
+    var loadOptions = function() {
+      var view_mode = $.QueryString["view"];
+      if (view_mode && (view_mode === "sidebyside" || view_mode === "code" || view_mode === "output")) {
+        $('#view_' + view_mode).click();
+      }
+
+      var genJS = $.QueryString["js"];
+      if (genJS) {
+        $('input:checkbox[name=genjs]').prop('checked', genJS == "true");
```
paf31: Use `===`?

soupi:
i added another checkbox for this: https://github.com/purescript/trypurescript/issues/14

but if that exists it begs a compile button.

so i thought about putting a checkbox next to a label that says compile, the compile label serves like a button to compile and not toggle auto reload and the checkbox toggles auto-reload;

but I'm not sure if that's reasonable and if it isn't too much.

I can also remove it if needed.

css/index.css
```diff
@@ -18,8 +18,78 @@ body {
...
+  vertical-align: bottom;
+}
+#showjs, #compile {
+  margin-left: 10px;
+}
+#auto_compile {
+  margin-left: 10px;
+}
+
+#view_mode {
```
paf31: Let's add `cursor: hand` here.

js/index.js
```diff
@@ -12,6 +12,22 @@
...
+        $('#view_' + view_mode).click();
+      }
+
+      var showjs = $.QueryString["js"];
+      if (showjs) {
+        $('input:checkbox[name=showjs]').prop('checked', showjs === "true");
+      }
+      var auto_compile = $.QueryString["compile"];
+      if (auto_compile) {
+        $('input:checkbox[name=auto_compile]').prop('checked', auto_compile === "true");
```
paf31: You can use `.checked(...)` here.

paf31:
This looks good generally, but the layout is wrong on mobile (I tried on iOS). I think you might need to experiment with the wrapping/height options.

soupi:
Alright, I'll take a look tonight or tomorrow night.

paf31:
Looks great! Thanks very much.

texastoland:
Wow @soupi very nice!

PR #24, opened 2016-07-10
Add files for Pux, Flare and Halogen

paf31:
@garyb @alexmingoia @sharkdp Could you please review these?

I will deploy these today, and share the API links. You will need to bundle the output for your client applications using `psc-bundle` and the `bower.json` file from the appropriate directory under `staging`.


paf31:
To be clear, by "review these", I meant the `bower.json` files 😄

paf31:
Ok, these are now live at 

- https://compile.purescript.org/flare2 (I'll change over the URL and remove the old service once the client has been updated)
- https://compile.purescript.org/pux
- https://compile.purescript.org/halogen

sharkdp:
Thank you very much!!! :+1: 

I was trying to update the client yesterday, but experienced some problems. I will report back when I get it up an running. Could take a few days, as I don't have much time right now..

PR #27, opened 2016-08-26
Save code buffer

### LINKED ISSUE #12: Save code buffer
Several times I've refreshed the page (<kbd>Super+R</kbd>) instead of opening a new tab (<kbd>Super+T</kbd>) 😭
- Easy fix: ask whether to navigate away from an edited buffer.
- Nice fix: keep active buffer in local storage.

matthieubulte:
Should close #12 

The idea here is to have a session per tab running updating the localSorage to preserve the code in case of reload or browser crash.

### DIFF
--- a/js/index.js
+++ b/js/index.js
@@ -8,7 +8,17 @@
             b[p[0]] = decodeURIComponent(p[1].replace(/\+/g, " "));
         }
         return b;
-    })(window.location.search.substr(1).split('&'))
+    })(window.location.search.substr(1).split('&'));
+
+    $.setQueryParameters = function(params) {
+        var url = location.href.split('?')[0];
+        var encodedParams = Object.keys(params).map(function(key) {
+            return key + '=' + encodeURIComponent(params[key]);
+        }).join('&');
+
+        document.location = url + '?' + encodedParams;
+    };
+
 })(jQuery);
 
 $(function() {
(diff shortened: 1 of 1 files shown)


js/index.js
```diff
@@ -48,6 +95,7 @@ $(function() {
             $('#' + ta_name).val(session.getValue());
             if ($("#auto_compile").is(":checked")) {
               compile();
+              cacheCurrentCode();
```
paf31: Here, we only save if the compile checkbox is set. Maybe we should save anyway.

js/index.js
```diff
@@ -28,6 +36,45 @@ $(function() {
...
+            return Math.floor((1 + Math.random()) * 0x10000)
+              .toString(16)
+              .substring(1);
+          }
+          return s4() + s4() + '-' + s4() + '-' + s4() + '-' +
+            s4() + '-' + s4() + s4() + s4();
+        }
+
+        var sessionId = $.QueryString['session'];
+        if (!sessionId) {
```
paf31: If the string is empty, this will take the `if` branch, which is fine, but `addQueryStringField` then results in `&session=&session=...` (minor point)

paf31:
Looks great!

matthieubulte:
@paf31 updated the code to handle the issues you pointed.

paf31:
Thanks!

paf31:
After merging, the editor doesn't show up for me any more. Could you please take a look? The old version was working fine for me.

PR #28, opened 2016-11-08
0.10 Updates

paf31:
This isn't complete yet. I still need to:

- [x] Update the license file
- [x] Use the published version of `purescript`
- [ ] Update the sample files

This changes things so that we build from source rather than requiring externs to be built already. This means it should be simpler to set up things like Try Halogen later.

I wonder if we should just distribute this with the compiler actually...


paf31:
This now bundles JS files on the server too, which should make spinning up custom servers even simpler. You can just point the browser to `/bundle`.

paf31:
I'll deploy this over the weekend hopefully, if someone could review it and the other open PR before then (please).

PR #34, opened 2016-12-02
Adding a button to publish code as an annonymous gist

soupi:
Adds a 'Share' button at the top right of the screen.

Pressing on it will trigger a confirm message. after that it will:
- Create an annonymous gist via github
- Refresh the page with the `gist` query parameter updated to the newly created gist
or it will fail and `alert` the user and will print the response object in console.

A few more changes:

- Tooltips for buttons
- pointer cursor for buttons
- Don't load session if `gist` query parameter is present (does that make sense?)

[Viewable here](https://soupi.github.io/trypurescript).

### DIFF
--- a/js/index.js
+++ b/js/index.js
@@ -236,19 +236,15 @@ $(function() {
     };
 
     var loadFromGist = function(id) {
-
         $.ajax({
             url: 'https://api.github.com/gists/' + id,
             dataType: 'json'
         }).done(function(gistInfo) {
-
             tryLoadFileFromGist(gistInfo, "Main.purs")
                 .done(function(code) {
-
                     code && $('#code_textarea').val(code);
                     setupEditor();
                 }).fail(function() {
-
                     console.log("Unable to load gist contents");
                     setupEditor();
                 });
@@ -258,7 +254,6 @@ $(function() {
             setupEditor();
         });
     };
-
 
     $('#showjs').change(compile);
     $('#compile_label').click(compile);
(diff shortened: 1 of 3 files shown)


js/index.js
```diff
@@ -285,16 +280,45 @@ $(function() {
         $('#showjs').show();
       }
     });
+    var publishNewGist = function() {
+        if (!confirm('Do you really want to publish as a Gist?')) { return; }
```
paf31: Maybe add something here along the lines of "This code will be publicly visible"? And add the word "anonymous" before Gist so it's clear it's not associated to the user's account if they happen to be logged in too?

js/index.js
```diff
@@ -285,16 +280,45 @@ $(function() {
...
+            data: JSON.stringify(data)
+        })
+            .success( function(e) {
+                console.log(e);
+                delete $.QueryString.session;
+                $.QueryString.gist = e.id;
+                $.setQueryParameters($.QueryString);
+            })
+            .error( function(e) {
+                console.warn("Gist creation failed: ", e);
```
paf31: Maybe this should alert the user too (without the `e` part of course) and direct them to look at the console output for more info?

paf31:
👍 Looks great, thanks!

paf31:
By the way, on my screen, it wasn't obvious that share was there, until I looked for it and found it on the right. Why not keep it with the other buttons?

js/index.js
soupi: sure. sounds good.

soupi:
I didn't like the way it looked much, especially since the Show JS can appear and disappear causing it to move. for the case we add more interaction with gists I thought it'd be better to group them to the right.
Also, it will probably look better after adding a few icons :)

soupi:
I made the following changes:

- Added an alert on failure
- Edited the text on gist creation
- Moved the share button before the compile button. does that looks better?

paf31:
Looks great, thanks!

PR #36, opened 2016-12-03
adding a readme and a help button linking to it

soupi:
This time the help button is on the top right. Will it go through? ☺


paf31:
Yes, thanks! 😄

PR #37, opened 2016-12-03
New layout + backend

soupi:
- Adding a drop down menu for view mode and backends
- Adding a layout for mobile and small screens with a hamburger menu
- Adding multiple backends support. for now only thermite
- Adding JS deps for thermite (react and friends)

[Preview](https://soupi.github.io/trypurescript)


js/index.js
```diff
@@ -128,19 +137,41 @@ $(function() {
...
+            iframe.write(
+              [ '<!DOCTYPE html>'
+              , '<html>'
+              , '  <head>'
+              , '    <title>Try Thermite</title>'
+              , '    <link rel="stylesheet" href="//maxcdn.bootstrapcdn.com/bootstrap/3.3.6/css/bootstrap.min.css">'
+              , '    <link rel="stylesheet" href="css/style.css">'
+              , '  </head>'
+              , '  <body>'
+              , '    <div id="console"></div>'
```
paf31: I'm not sure if we really need this virtual console for Try Thermite. I'd be fine with going back to using the regular console actually.

js/index.js
```diff
@@ -128,19 +137,41 @@ $(function() {
...
-            , '    <link rel="stylesheet" href="css/style.css">'
-            , '  </head>'
-            , '  <body>'
-            , '    <div id="console"></div>'
-            , '  </body>'
-            , '</html>'
-            ].join('\n')
-        );
+        if ($('input[name=backend_inputs]').filter(':checked').val() === "thermite") {
+            iframe.write(
```
paf31: A lot of this is common. Maybe we can extract the common bits?

js/index.js
```diff
@@ -196,18 +217,36 @@ $(function() {
...
                     } else {
-                      $.get(endpoint + '/bundle').done(function(bundle) {
-
-                          execute(res.js, bundle);
-                      }).fail(function(err) {
-
-                          console.log("Unable to load JS bundle");
-                      });
+                      if ($('input[name=backend_inputs]').filter(':checked').val() === "thermite") {
+                           $.ajaxSetup({ dataType: 'text' });
```
paf31: You only need to run this once, when the application starts.

paf31:
Looks great, especially on mobile now!

My main request is just that we should load the Try Thermite sample code if the user switches to the Thermite backend, either using the dropdown or on initial load in the URL.

soupi:
Right. thanks. I'll keep working on it :)

paf31:
This works nicely now, but I spotted one issue: clicking Share while using the Thermite backend changes back to the core libraries backend.

Also, do you think it makes sense to call it "Core" or "Core libraries" instead of "PureScript"?

soupi:
Yeah, "Core" sounds good. Thanks for spotting the back. i'll fix it now.

paf31:
👍 Great, thanks!

PR #39, opened 2016-12-08
Adding slides as a possible backend

soupi:
This depends on a backend for `purescript-slides` at endpoint `https://compile.purescript.org/slides` (can be changed of course) that has packages as suggested in [this bower.json](https://gist.github.com/soupi/aec3c558ac1fedce6c4b2e740dc3e9e0).

[Preview](https://soupi.github.io/trypurescript) will not work until we have a backend of course :)

### DIFF
--- a/js/index.js
+++ b/js/index.js
@@ -42,6 +42,8 @@ var coreStart =
     ,'      , h2 (text "Try PureScript Libraries")'
     ,'      , list [ link "?backend=thermite" (text "Try Thermite") '
     ,'               <> text ", a front-end library for PureScript which uses React"'
+    ,'             , link "?backend=slides" (text "Try Slides") '
+    ,'               <> text ", an EDSL in PureScript for creating presentations"'
     ,'             ]'
     ,'      , h2 (text "Share Your Code")'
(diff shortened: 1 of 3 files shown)


js/index.js
```diff
@@ -179,6 +181,75 @@ thermiteStart =
     ,'main = T.defaultMain spec unit'
     ].join('\n');
 
+var slidesStart =
```
paf31: I really think we ought to download these when the application starts.

paf31:
Looks good, I'll deploy this later, thanks!

js/index.js
```diff
@@ -179,6 +181,75 @@ thermiteStart =
...
+    ,'      , center $'
+    ,'          text "(In "'
+    ,'          <+> link "http://purescript.org" (text "PureScript")'
+    ,'          <>  text ", Using "'
+    ,'          <+> link "https://github.com/soupi/purescript-slides" (text "purescript-slides")'
+    ,'          <>  text ")"'
+    ,'      , text ""'
+    ,'      , center $'
+    ,'          text "New to PureScript? Perhaps you want to"'
+    ,'          <+> link "?backend&core" (text "try PureScript")'
```
paf31: `&` should be `=` here, but I can fix this.

paf31:
Deployed, thanks!

PR #43, opened 2016-12-09
Add client code for try-flare.

sharkdp:
I have adapted my old Try Flare client code to the new interface with just two panes. Thanks again for updating the backend!


paf31:
That was quick 😄 

Thanks!

sharkdp:
Very cool, Thanks!

PR #46, opened 2016-12-10
Increase label width and fix Firefox invisible slider

### LINKED ISSUE #44: Flare CSS problems
@sharkdp  First of all I wanted to say flare still looks awesome!

I have noticed a few CSS problems in the try purescript site:

1. in the temperature example the temperature picker goes over the temperature label (probably because the width for it is 100px). Perhaps a putting the components in a table will work?
2. Components like the font size for the title or hue and radius in the drawing example are invisible to me. I can manipulate them if i press where they are supposed to be but i can't see them.

This is using Firefox 50.

edit: I can, however, see the sliders for hue and radius fine on my phone (also firefox).

sharkdp:
This fixes #44

### DIFF
--- a/css/flare.css
+++ b/css/flare.css
@@ -16,7 +16,7 @@ a, a:visited, a:active, a:hover {
 
 .flare-input > label {
     display: inline-block;
-    width: 100px;
+    width: 150px;
 }
 
 .flare-input {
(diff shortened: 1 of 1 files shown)


paf31:
Thanks!

PR #47, opened 2016-12-10
adding meta viewport tag and fixing backend&core in startSlides

soupi:
[Preview](http://soupi.github.io/trypurescript) if there's a need for that :)


paf31:
Thanks!

PR #48, opened 2016-12-10
meta content tag and image from imgur over https

soupi:
[Preview](https://soupi.github.io/trypurescript) as usual.


paf31:
Thanks!

PR #57, opened 2017-01-20
Add try-mathbox

rintcius:
- add try-mathbox
- allow first `Main.purs` to be loaded from gist (instead of just snippet that's part of `index.js`)
    IIRC I saw a comment from Phil somewhere that adding this would be good. Advantage for me is that I can improve the demo without having to change anything inside trypurescript. So this PR seemed like a good opportunity to implement this feature, but happy to change if preferred.
- refactored `iframe` creation a bit because old setup didn't work with mathbox integration.
  New setup should work with all backends.


js/index.js
```diff
@@ -635,6 +659,20 @@ $(function() {
...
                           myconsole.warn("Unable to load JS bundle", err);
                         });
+                      } else if ($('input[name=backend_inputs]').filter(':checked').val() === "mathbox") {
+                           $.when($.get("js/mathbox-bundle.js"),
+                                  $.get(backend.endpoint + "/bundle")
+                                ).done(function(mathboxBundle, bundle) {
+
+                           var replaced = bundle[0].replace(/require\("Data.Either"\)/g, "PS['Data.Either']")
+                                                   .replace(/require\("Data.Maybe"\)/g, "PS['Data.Maybe']")
+                                                   .replace(/require\("Mathbox.Field"\)/g, "PS['Mathbox.Field']")
```
rintcius: Not sure why this is needed exactly. 

When similar examples are created [here](https://github.com/rintcius/purescript-mathbox/blob/master/build-examples.sh) then the bundle works as-is.

soupi:
I suggest that if we are changing the Main.purs snippets mechanism we might as well change them all so we won't have to deal with 'undefined's and stuff in the code. what do you think @paf31 @sharkdp ?

I've added a gist for the purescript-slides snippet as well: [c62b5778a6a5f2bcd32dd97b294c068a](https://gist.github.com/soupi/c62b5778a6a5f2bcd32dd97b294c068a).

paf31:
This looks great!

paf31:
I agree with regards to the initial snippets though.

rintcius:
Indeed, I had in mind to get rid of the initial snippet as well. Is it ok with you guys to do it in a separate PR though?

paf31:
Yes, I think that sounds good.

js/index.js
paf31: This definitely shouldn't be necessary.

rintcius: Good to mention, I have used these `require`'s explicitly in `js`:
https://github.com/rintcius/purescript-mathbox/blob/master/src/Mathbox/Mathbox.js#L6-L8

Could this be due to a mismatch in what `pulp browserify` does & how this bundle was created?

paf31: Well the server uses `psc-bundle`, which explains the issue. Is it possible to remove those requires from the library?

rintcius: @paf31 ok, I managed to get rid of them and published a new version: https://pursuit.purescript.org/packages/purescript-mathbox/0.1.1

If you can make it available at the server side, then I will cleanup the code here.

BTW Wouldn't it be better if `psc-bundle` handled it the same as `pulp browserify`?

paf31: Thanks.

> Wouldn't it be better if psc-bundle handled it the same as pulp browserify?

Maybe, but that would mean running `browserify` on the server, which I'd rather not do.

paf31: Updated, thanks!

rintcius: Ok, I didn't mean to run `browserify` on the server but rather fix it in `psc-bundle` (can imagine that that's also something you'd rather though - I don't know anything about the internals of `psc-bundle`)

rintcius: Awesome!

paf31: I'd prefer the solution which involves `psc-bundle` doing the least amount of work, since it's going to be changed quite a lot before 1.0.

rintcius:
Pushed the clean up

paf31:
👍 Looks good, thanks!

soupi:
@rintcius mathbox looks really cool! We need to add a link to it in the core initial snippet as it looks like its missing.

rintcius:
Thank you guys - happy to see it on try-purescript!

@soupi  Thanks for spotting - will add the link when getting rid of the initial snippets

PR #59, opened 2017-01-26
Remove initial snippets

rintcius:
This PR gets rid of the initial snippets.

I have split it up in separate commits, meant to clear up (a bit) why I have changed them. Feel free to ask if something is not clear enough.

### DIFF
--- a/js/index.js
+++ b/js/index.js
@@ -420,11 +420,13 @@ $(function() {
                     setupEditor();
                 }).fail(function() {
                     console.log("Unable to load gist contents");
+                    $('#code_textarea').val("-- Unable to load gist contents. Gist = " + id);
                     setupEditor();
                 });
         }).fail(function() {
 
             console.log("Unable to load gist metadata");
+            $('#code_textarea').val("-- Unable to load gist metadata. Gist = " + id);
             setupEditor();
         });
     };
--- a/js/index.js
+++ b/js/index.js
@@ -25,314 +25,27 @@ $.ajaxSetup({ dataType: 'text' });
 
 })(jQuery);
 
-
-var coreStart =
-    ['module Main where'
-    ,''
-    ,'import Prelude'
-    ,'import Data.Foldable (fold)'
-    ,'import TryPureScript'
-    ,''
-    ,'main ='
-    ,'    render $ fold'
-    ,'      [ h1 (text "Try PureScript!")'
-    ,'      , p (text "Try out the examples below, or create your own!")'
-    ,'      , h2 (text "Examples")'
-    ,'      , list (map fromExample examples)'
-    ,'      , h2 (text "Try PureScript Libraries")'
-    ,'      , list [ link "?backend=thermite" (text "Try Thermite") '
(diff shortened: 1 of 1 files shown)


js/index.js
```diff
@@ -711,11 +420,13 @@ $(function() {
                     setupEditor();
                 }).fail(function() {
                     console.log("Unable to load gist contents");
+                    $('#code_textarea').val("-- Unable to load gist contents. Gist = " + id);
```
paf31: I think we should just use `console.error` for this.

paf31:
👍 Looks great, thanks!

js/index.js
rintcius: I added it to make sure that it's not retrying in case of failure (and hitting github's ratelimit quickly as I found out)

paf31: Why does it have to retry?

rintcius: I doesn't have to - just meant that it currently does, because of https://github.com/purescript/trypurescript/blob/gh-pages/js/index.js#L375

I can try and make this behaviour more explicit if you prefer (e.g. adding a `isRetry` flag or something to the functions)

paf31: Ok, I think we should try to break down `loadOptions` and/or `setupEditor` a bit, so that we don't end up looping, in that case.

rintcius: Solution I see without restructuring things too much is to parameterize

```
var loadOptions = function(tryLoad) { ...

var setupEditor = function(tryLoad) { ...
```

Then this logic only needs to change:
https://github.com/rintcius/trypurescript/blob/dcb8522982f2ca0fd22e3f7a2177dec271ef244e/js/index.js#L88-L90

into 

```
      if (tryLoad && $('#code_textarea').val() === "") {
          loadFromGist(backend.mainGist);
      }
```

I think splitting up would mean taking the `if` statement above out of `loadOptions` and into its own function but I think this would end up to be more complex than parameterizing (because AFAICS the only change we'd want in the flow is this `if` statement)

paf31: Sounds good to me!

rintcius:
Just noticed this warning in the console because of the `async: true`'s that I added:

```
query.js:8672 Synchronous XMLHttpRequest on the main thread is deprecated because of its detrimental effects to the end user's experience. For more help, check https://xhr.spec.whatwg.org/.
```

User experience looks fine though while testing. When I have some spare time again, I could look into getting rid of those `async: true`'s (looks like we then need some more restructuring after all).

paf31:
Why do we need to use synchronous ajax?

rintcius:
It would be better without, but I added them because otherwise it can ask for confirmation multiple times when you change the backend via the menu.

paf31:
I don't want to have to resort to using synchronous calls, so I think we should figure out why that's happening and fix it instead. I can take a look if you don't have time though.

rintcius:
Yes, would be better to fix properly. Feel free to pick up from here.

paf31:
Finished in #60

PR #65, opened 2017-02-20
Fix github unicorn error and cache

soupi:
This fixes:

1. Github's unicorn error when trying to "show js" or press the compile button
2. Fix cache string - will now try to use the cache string to load previously cached code and backend settings

Changes can be viewed [here](https://soupi.github.io/trypurescript).

### DIFF
--- a/js/index.js
+++ b/js/index.js
@@ -174,12 +174,13 @@ $(function() {
       if (code) {
         $('#code_textarea').val(code);
       }
+      return backend;
     }
   }
 
   var editor, cleanupActions = [];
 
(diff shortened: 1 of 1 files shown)


paf31:
Thanks!

paf31:
Actually now I always get an empty input on load.

soupi:
That's weird, it works for me on computer and mobile, can you try a hard refresh or something?

soupi:
It appears that what you describe indeed happen on chromium (not on firefox which i use) so now i have a way to replicate. Will work on this tonight.

PR #66, opened 2017-02-21
Revert "Fix github unicorn error and cache"

paf31:
Reverts purescript/trypurescript#65


paf31:
@soupi I'll revert this for now, since it seems to break things.

PR #67, opened 2017-02-21
Fixing PR #66

soupi:
This should fix #66 That was introduced in #65 for chromium browser

Please check that it works [here](http://soupi.github.io/trypurescript) and tell me if it isn't.

Thanks

### DIFF
--- a/js/index.js
+++ b/js/index.js
@@ -173,8 +173,8 @@ $(function() {
       }
       if (code) {
         $('#code_textarea').val(code);
+        return backend;
       }
-      return backend;
     }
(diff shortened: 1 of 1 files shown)


paf31:
Looks good, what was the issue?

soupi:
I'm not entirely sure tbh, I guessed that there was a scenario where the backend was saved somehow but the code wasn't and i decided to make sure to only returned the backend string if the code was set. Ran some tests and it seems to work.

paf31:
Thanks!

PR #68, opened 2017-02-25
favicon + a few fixes

soupi:
* add favicon
* replace .load() (deprecated) with .ready()
* add defaultBundleAndExecute for each backend
* avoid loading twice when there's no session (firefox)
* remove '/' in query parameter names (may happen when setting location.href)
* update jquery version


paf31:
Thanks, looks great!

PR #70, opened 2017-02-25
busy wait on body for iframe to load

soupi:
In the previous fix of `.load()` -> `.ready()` it appear that now `iframe` is `ready` before body is ready. this is ok for most cases but not for mathbox which has an extra js dep. and at the point of `var body` the `iframe` is not ready yet.

This fixes this issue by busy waiting on `body` to be ready. It's a bit ugly, but i haven't found a better solution from googling this issue.

### DIFF
--- a/js/index.js
+++ b/js/index.js
@@ -276,18 +276,17 @@ $(function() {
     script.appendChild(iframe.createTextNode(scripts));
 
     $('iframe').ready(function() {
+      var checkExists = setInterval(function() {
+        var body = iframe.getElementsByTagName('body')[0];
+        if (body) {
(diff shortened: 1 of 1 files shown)


paf31:
Did this work with `load` then? But `load` is now deprecated?

soupi:
`load` only works in some browsers apparently, for example it does not work on my android browser at all. Initially @naomik reported a bug that arose from using `load` and I think she used chrome on osx?

I'd really like to use something like `load`, but one that works, instead :(

paf31:
Okay, let's go with this for now then. Thanks!

PR #72, opened 2017-03-15
Updating slides css for code elements

soupi:
- [before](http://try.purescript.org/?backend=slides) (go to 2nd slide)
- [after](http://soupi.github.io/trypurescript/?backend=slides) (go to 2nd slide)

Thanks


paf31:
Thanks!

PR #74, opened 2017-03-31
Updates for 0.11

paf31:
My plan is to deploy this over the weekend. I will start by disabling all "Try X" services, and I'll turn them back on as soon as the necessary libraries are updated and I have a working `psc-package.json` or `bower.json` for each one.

So there's no real rush, since the core "Try PureScript" will be working, but the other services will be offline initially.

cc @sharkdp @soupi @rintcius

### DIFF
--- a/server/Main.hs
+++ b/server/Main.hs
@@ -32,6 +32,7 @@ import           GHC.Generics (Generic)
 import qualified Language.PureScript as P
 import qualified Language.PureScript.Bundle as Bundle
 import qualified Language.PureScript.CodeGen.JS as J
+import qualified Language.PureScript.CodeGen.JS.Printer as P
 import qualified Language.PureScript.CoreFn as CF
 import qualified Language.PureScript.Errors.JSON as P
 import qualified Language.PureScript.Interactive as I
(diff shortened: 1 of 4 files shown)


soupi:
Turning other services is a bit of a shame imo... it would be nice if they could co-exist. but if that the plan then alright.

paf31:
Well, I can leave them running as-is for now.

rintcius:
Yes, maybe good to give a bit of time for the ecosystem to catch up with 0.11?

BTW how long does it usually take for libs to catch up?

paf31:
purescript-contrib is usually updated relatively quickly. What else do you depend on?

rintcius:
There were a couple of libs that were not part of the package set yet (so these are probably not that actively maintained), but I can upgrade them as well if necessary.

paf31:
The main service is now running 0.11.1, while the other services continue to run 0.10.7. Please let me know when your libraries are ready and I can update them. Thanks!

paf31:
Try Thermite is now running latest again.

rintcius:
mathbox 0.4.0 is now on purescript 0.11 - these are the deps https://github.com/rintcius/purescript-mathbox/blob/master/bower.json

ping me when try-mathbox is running against 0.11 & I'll update the try-mathbox gists to 0.11 too

paf31:
@rintcius Thanks! Do you think it would be much work to update the package set, and specify the dependencies that way? I'd like to transition to having Try PureScript use `psc-package` and keeping the dependencies in the `staging/` directory.

rintcius:
@paf31 Ok, I'll have a look!

paf31:
Ping @sharkdp @soupi - do you think you will be able to update your packages soon?

soupi:
Hopefully this evening

sharkdp:
Thanks for the reminder!

In the meantime, all necessary libraries for try-flare have been updated to 0.11, but I'm not sure how this works with the package sets now.. should we send PRs for https://github.com/purescript/package-sets ?

paf31:
Yes, I think the simplest way would be to update the main package set, if possible, and then provide a `psc-package.json` file with the dependencies I need to compile things. It should be possible to use a previously-tagged package set in order to save some time hunting down all transitive dependencies.

sharkdp:
This would mean that smolder, flare and flarecheck have to be added to the global package-set. Is this really intended? Sorry for asking, but I didn't fully get all the concepts behind psc-package/package-sets.

paf31:
Yes, we'd need to add those. Adding a package doesn't imply any sort of ongoing maintenance, just that the package will not break semantic versioning going forward, so there is no harm in adding a package to the set if it works.

Issue #77, opened 2017-06-13
Update compiler version


matthewleon:
would be nice to have this on 0.11.5. Was trying to write some code with InstanceSigs.

paf31:
Core, Thermite, and Behaviors backends currently use 0.11.6. The other backends need to be updated, but that's a library issue (the Bower/`psc-package` config needs to be updated).

PR #78, opened 2017-06-13
update purescript dep to purs-0.11.5

matthewleon:
addresses https://github.com/purescript/trypurescript/issues/77

This compiles for me. Not sure if there is any other testing I should do?

### DIFF
--- a/stack.yaml
+++ b/stack.yaml
@@ -3,5 +3,5 @@ flags: {}
 packages:
 - '.'
 extra-deps:
-- purescript-0.11.2
+- purescript-0.11.5
 - pipes-http-1.0.5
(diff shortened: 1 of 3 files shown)


paf31:
We should probably also update the core library versions, but that's not a big deal.

Thanks!

matthewleon:
> We should probably also update the core library versions

FWIW I've gone ahead and updated the Stack LTS and one of the fixed dep versions.

paf31:
Oh, sorry, I meant the PureScript library dependencies, but it's good to update these too.

paf31:
This is the line which needs to be updated:

https://github.com/purescript/trypurescript/blob/eaecf651c6a311e771ed025053cb08fa5a5b1f67/staging/core/psc-package.json#L3

matthewleon:
> This is the line which needs to be updated:

Done. Thanks.

PR #80, opened 2017-06-25
adding a link to gist under view mode

soupi:
This commit adds a link to a gist on github (useful for mobile users who just want to read the code) as suggested [here](https://twitter.com/bkase_/status/879030876377665536).

- When a link to gist is available it will add a 'Gist' link under 'View Mode' menu ([Preview](https://soupi.github.io/trypurescript/))
- When  a gist is not available it will not show this option at all ([Preview](https://soupi.github.io/trypurescript/?gist=3f735aa2a652af592101))

### DIFF
--- a/css/index.css
+++ b/css/index.css
@@ -98,6 +98,13 @@ body {
   border-left: solid 1px #fff;
   border-right: solid 1px #fff;
 }
+#view_mode a {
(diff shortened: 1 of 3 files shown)


paf31:
Great, thanks!

PR #88, opened 2017-08-12
Added tryHalogen backend

sudhirvkumar:
@paf31 Sending you the PR as discussed in IRC

### DIFF
--- a/js/index.js
+++ b/js/index.js
@@ -99,6 +99,15 @@ $(function() {
         extra_body: '<canvas id="canvas" width="800" height="600"></canvas>',
         bundleAndExecute: defaultBundleAndExecute
       };
+    } else if (backend === "halogen") {
+      return {
+        backend: "halogen",
+        endpoint: "https://compile.purescript.org/halogen",
+        mainGist: "15536d7b4c66ac512b2c8dca98e0e3d3",
+        extra_styling: '',
+        extra_body: '',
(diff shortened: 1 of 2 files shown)


js/index.js
```diff
@@ -287,23 +275,29 @@ $(function() {
 
     var scripts = [bundle, wrapped].join("\n");
 
-    var script = iframe.createElement('script');
```
paf31: Did you test this on other backends? I seem to remember it was important.

sudhirvkumar: I checked every single backend available in try.purescript.org just now... (twice) and everything is working... including Halogen

sudhirvkumar: Having said that I tested only in Chrome!

paf31:
Could you please rebase this? I've added a `DEPENDENCIES` file which you could regenerate in order to give attribution here.

Thanks!

sudhirvkumar:
@paf31 sure will do

RAbraham:
Thanks @sudhirvkumar for working on this. Excited to play with it.

rintcius:
@sudhirvkumar let me know if you don't have time to finish this. I'm happy to help

hdgarrood:
Thanks for the PR, although I'm going to close it as we are now using a different design where different backends are no longer a thing, see #128. Getting Halogen examples working with the new design will now hopefully just be a case of updating the `psc-package.json` file to include it.

PR #90, opened 2017-08-22
PureScript rewrite

### LINKED ISSUE #62: Port to PureScript
I think we have enough JS code here now that we should port it to PureScript.

paf31:
Closes #62.

I started rewriting things in small steps, as an exercise, instead of rewriting everything at once. This way, we'll at least start getting some benefit from the type system.

More to come...


paf31:
This is now ready for review. I might refactor the jQuery code a bit, but I'll probably leave that for a separate PR, since everything is working here.

paf31:
I'll merge this soon if there are no objections. I've tested it pretty thoroughly.

@soupi What do you think?

soupi:
I'll take a good look at this this evening :)

src/Main.purs
```diff
@@ -0,0 +1,529 @@
...
+    JQuery.empty jq
+    JQuery.create "<div>" >>= \div -> do
+      JQuery.addClass "loading" div
+      JQuery.appendText "Loading..." div
+      JQuery.append div jq
+
+  code <- fold <$> (JQuery.select "#code_textarea" >>= JQuery.getValueMaybe)
+
+  let displayPlainText s =
+        JQuery.select "#column2" >>= \jq -> do
```
soupi: why not use do notation?

src/Main.purs
```diff
@@ -0,0 +1,529 @@
...
+            String
+            String
+            (EffFn1 (dom :: DOM | eff) JQuery.JQuery Unit)
+            Unit
+
+-- | Execute the compiled code in a new iframe.
+execute :: forall eff. JS -> JS -> BackendConfig -> Eff (dom :: DOM | eff) Unit
+execute js bundle bc@(BackendConfig backend) = do
+  let html = joinWith "\n"
+        [ "<!DOCTYPE html>"
```
soupi: This could be a bit more readable and easier to change with `"""` instead

src/Main.purs
```diff
@@ -0,0 +1,529 @@
...
+
+    ok <- window >>= confirm ("Replace your current code with the " <> newBackend.backend <> " backend sample code?")
+    if ok
+      then navigateTo ("?backend=" <> newBackend.backend)
+      else void $ setTimeout 1000 do
+             compile bc_
+             cacheCurrentCode
+    hideMenus
+
+getBackend :: Backend -> BackendConfig
```
soupi: This should go in `Types.purs` imo.

paf31:
Thanks for the review @soupi, I'll get back to this today hopefully.

src/Main.purs
paf31: Good point, thanks.

src/Main.purs
paf31: Done, thanks.

src/Main.purs
paf31: I moved all of this under API actually. To do what I wanted, it had to be that way to avoid a dependency cycle.

paf31:
I've refactored the view code a little bit. I think it's slightly nicer now, but we could still do some more work on it.

soupi:
Congrats 🎉

PR #93, opened 2017-08-31
Fix a parser error when an error position is missing

paf31:


soupi:
maybe it's best to untrack `js/index.js` and add it to .gitignore ?

paf31:
But we need it there for the site to work.

soupi:
ah! yeah... ^_^"

PR #102, opened 2017-09-04
Pure CSS menu

### LINKED ISSUE #98: Show an appropriate error on mobile devices
And redirect to the gist, if appropriate.

paf31:
This changes things a bit so that we rely less on jQuery and more on CSS for certain visual elements.

It also fixes #98 by displaying a message and hiding the code editor by default on small screens.

## Desktop View

screen shot 2017-09-04 at 3 19 51 pm
screen shot 2017-09-04 at 3 20 09 pm

## Small Screen (< 720px) View

screen shot 2017-09-04 at 3 20 37 pm

## Small Screen with Gist link

screen shot 2017-09-04 at 3 22 53 pm


paf31:
@soupi Could you please review this if you have time?

paf31:
@soupi I'll merge this, but if you have concerns, let me know.

soupi:
I'm sorry for the late response!

I really like the new look and feel! I do have a few comments though:

- Editing code in mobile - while not having a very good experience is still quite useful and I often use it when i want to write short examples or answer questions. Are we sure we want to disable this functionality?

A few small things:

- The drop menu's items can be highlighted and one must press the text and not the whole line

P.S. the logo animation is cool 👍

paf31:
Thanks @soupi - I just fixed the drop down menu issue.

Yes, I'm not pleased about code editing on mobile, but the problem is that Ace just really doesn't work well on iOS. I'm not sure about other platforms. What are you using?

soupi:
I'm on android using either firefox or brave. it's not good but somewhat manageable. for example if i disabled auto compilation and scroll on the edge of the screen in code mode i can edit code.

I wonder if anyone else uses try-purs on mobile other than viewing compiled code?

PR #105, opened 2017-09-09
add try-mathbox psc-package (package set psc-0.11.6)

rintcius:


rintcius:
Client part in #104

paf31:
Thanks!

paf31:
Deployed, thanks!

@joneshf @sharkdp You might find this useful as an example. I'll try to improve the docs as well.

joneshf:
So, just add a psc-package? Does anything need to be deleted?

paf31:
You would need to

- find or create a package set with the version of flare you want
- make a local psc-package config and compile it
- pass the source files to the trypurescript executable and test locally
- update the gist id in the client to point to one which works with that version of flare (I believe the current ones are under @sharkdp's account)

PR #109, opened 2017-09-10
Add a section for development to README

rintcius:


rintcius:
BTW I have added `npm run bundle` but the `open index.html` part doesn't work for me (because it's opening `file://` instead of `http://`). Anything I I'm missing?

sharkdp:
> but the open index.html part doesn't work for me

Same for me. I think it should be removed from `package.json`.

paf31:
Fantastic, thanks!

I agree about removing the `index.html` command.

rintcius:
@paf31 did you mean to close it without merging?

paf31:
No, sorry.

PR #110, opened 2017-09-11
setQueryStrings should replace existing parameters

MonoidMusician:
`union` is left-biased: https://pursuit.purescript.org/packages/purescript-maps/3.5.0/docs/Data.StrMap#v:union

This should fix the issue where saving a gist would not update the url if the url already had a gist parameter. But I haven't tested it 😁

### DIFF
--- a/js/index.js
+++ b/js/index.js
@@ -5000,7 +5000,7 @@ var PS = {};
   var setQueryStrings = function (ss) {
       return function __do() {
           var v = getQueryParams();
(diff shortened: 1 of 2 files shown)


paf31:
Could you please rebuild the JS and I'll merge this in?

MonoidMusician:
Okay, I made the corresponding change.

paf31:
Thanks!

PR #122, opened 2018-05-11
Idea: host our own gists?

soupi:
I tried thinking about how to get us back to the good old days where we could share our snippets straight from try.purs and came up with the idea of implementing my own limited pastebin  ([here](https://github.com/soupi/plaste)) and an interface for the client and server of try.purs.

I'm not sure if this approach is a good one, but I had some time so I tried it. What do you think?

client relevant PR here: #123


hdgarrood:
I think I'd prefer to go the route of authenticating to GitHub so that we can save gists, personally. Especially given the fate of lpaste; I think running a service which allows hosting anonymously-uploaded things is not really very desirable.

soupi:
Sure, makes sense.

PR #123, opened 2018-05-11
Idea: host our own gists? (Client)

soupi:
visit #122 for more information

Note: it seems that for this to work I still need to serve the plastes over https.


src/Try/Plaste.purs
```diff
@@ -0,0 +1,41 @@
+
+module Try.Plaste
+  ( uploadPlaste
+  , getPlasteById
+  ) where
+
+-- | An abstract data type representing the data we get back from the Plaste API.
```
paf31: ?

soupi: whoops

PR #124, opened 2018-05-15
Add variant and run

gabejohnson:


paf31:
Sorry, these can't go in the core set, since they are not core libraries. I explained this on Slack, but long term, I would like to offer the option of building arbitrary code, like Ellie does for Elm, but the externs are kept in memory on the server now, which puts a bound on the number of libraries we can provide.

For now, you can always create a local server which supports these.

PR #126, opened 2018-11-09
Bump version to 0.12

natefaubion:


natefaubion:
I don't fully understand this failure. It appears to build/install fine locally with these constraints. Any ideas?

natefaubion:
Exit code 137 appears to be because it was kill -9ed, usually for using too many resources.

natefaubion:
I think core is the only backend that is buildable. Is there anything I should do on the API side to disable the others? Or is that purely a `github-pages` thing?

paf31:
In the interests of getting this released, I think we should focus only on the core package set.  It's a change in the client code.

natefaubion:
I've verified that the package-set for core installs and the TryPureScript module compiles. Is there anything else for the main branch before venturing into gh-pages?

natefaubion:
I've updated this to 0.12.1

blinky3713:
What's the status with this and #127, are we able to get a working release with 0.12.1 sometime soon?

hdgarrood:
I suspect the failure is due to there not being any output for too long leading to Travis terminating the build. I'd argue that the best thing to do for now is just to ignore the CI failure.

natefaubion:
I've made updates for 0.12.3.

PR #127, opened 2018-11-10
Frontend updates for 0.12

natefaubion:


natefaubion:
This appears to work overall. Calling the API and compiling works, however the iframe is yielding a `TryPureScript is undefined` error. Not sure if there's something I didn't setup correctly.

natefaubion:
I think that is because of the `require` changes to add `index.js`.

natefaubion:
The client is now working as expected.

src/Main.purs
```diff
@@ -211,7 +211,7 @@ execute js bundle bc@(BackendConfig backend) = do
         ]
 
       replaced = replace' (unsafeRegex """require\("[^"]*"\)""" global) (\s _ ->
-        "PS['" <> String.drop 12 (String.take (String.length s - 2) s) <> "']") (unwrap js)
+        "PS['" <> String.drop 12 (String.take (String.length s - 11) s) <> "']") (unwrap js)
```
hdgarrood: Bit of a nit but can we do something like `s # String.split (String.Pattern "/") # Array.index 1` here instead? That should be more robust to these sorts of changes (if this ever changes again) and it's also a little clearer, I think.

hdgarrood:
Resolved by #135

Issue #132, opened 2020-01-02
Put all source files in one branch


hdgarrood:
At the moment, the source files for the server are in the `master` branch, but the source files for the frontend are in the `gh-pages` branch (along with the compiled output which is actually used by GH Pages). This has a few drawbacks:

- It's more awkward to clone the repo and get the client side and server side working together locally than it potentially could be if all code were in one branch; you need to have two separate git checkouts, for instance
- It's harder to keep track of which versions of the client are supposed to work with which versions of the server (if all the code were in one branch, it would be safe to assume that the server and client at any given commit are supposed to work together, or we could at least aim for that) 
- We have to check in a ton of compiled JS code alongside the sources for the frontend, which makes PRs difficult to review
- It ties merges to deployments in a way that can be awkward -- for example, if a change to the client has passed review and could be merged but can't be deployed yet, say, because the server needs to be deployed first. Right now I would like to merge #127, but it would break Try PureScript unless we deployed the server at the same time.

I think the last one is the most significant, as I think the way this repo is set up currently is more or less incompatible with the approach we were describing at https://discourse.purescript.org/t/looking-for-new-maintainers-for-purescript-org/1014/25 where we want to be able to automate deployment as much as possible.

If we put the frontend source in `master` alongside the server, I think local development and deployment for Try PureScript could be made a bit simpler: we could have all of the source files for both frontend and backend live in the same branch, and have scripts for local development which compile and run of the app for you. We could then have a separate script which generates the compiled frontend files ready for serving and checks those in on the `gh-pages` branch, so that the `gh-pages` branch would only be used for compiled build artifacts and not source code. (Or alternatively we could just stop using GH pages entirely; the CDN probably isn't giving us all that much benefit since the frontend only needs to be loaded once, more or less. But that's something we can consider separately)

kritzcreek:
According to https://help.github.com/en/github/working-with-github-pages/configuring-a-publishing-source-for-your-github-pages-site#choosing-a-publishing-source we could also serve the frontend from the `docs/` folder on the master branch. That way you could have PRs that modify both against the same branch.

natefaubion:
> That way you could have PRs that modify both against the same branch.

I think we still want the ability to merge changes without redeploying them. I think gh-pages is helpful for that at least.

hdgarrood:
I'd really like to keep compiled stuff separate for the sake of diffs.

hdgarrood:
[...]

PR #135, opened 2020-02-18
All source files in one branch

gabejohnson:
See #132

We would like to keep all source files in a single branch and use `gh-pages` purely for hosting
build artifacts.

Going forward we would build the PS source and push it (along with the html, css, other js dependencies, README.md, etc.) to `gh-pages`. No edits would happen directly in `gh-pages`.

So far I've done the minimum to move the frontend files into the mainline branch and get the Core backend working with `v0.13.6`. Much of the work was already done by @natefaubion 
 in #127 which I've merged into this branch. The other backends still need updating and there are none of the scripts mentioned in https://github.com/purescript/trypurescript/issues/132#issue-544413955.

I've forked the Core gist and the example gists to which it points to get them working with `Effect` and pointed `Try.API` to that fork. Going forward it may make sense to have those gists live under the `purescript` org.

I'm open to adding more to this PR (backend updates, build scripts) but wanted to keep it minimal and spur discussion concerning what needs to be done to get https://try.purescript.org up-to-date.


client/js/index.js
```diff
@@ -0,0 +1,6231 @@
+// Generated by purs bundle 0.12.0
```
hdgarrood: Should this file be ignored?

hdgarrood:
I'm very much in favour of this PR minimal, by the way. :+1:

client/js/index.js
gabejohnson: I had just added a it to `.gitignore` right before opening the PR, but forgot to remove it. Fixed.

citizengabe:
Those all sound like reasonable changes. I think replacing the `thermite` backend with `react-basic` also makes sense while updating the other backends.

Since I incorporated the changes from https://github.com/purescript/trypurescript/pull/127, that can probably be closed as well.

gabejohnson:
Those all sound like reasonable changes. I think replacing the `thermite` backend with `react-basic` also makes sense while updating the other backends.

Since I incorporated the changes from https://github.com/purescript/trypurescript/pull/127, that can probably be closed as well.

natefaubion:
I don't think it's worth updating backends. I would much rather merge up https://github.com/purescript/trypurescript/pull/128 and obviate the need for backends.

gabejohnson:
> I would much rather merge up #128 and obviate the need for backends.

That makes sense. I guess just updating some gists then.

PR #136, opened 2020-03-08
Merge changes from #128

gabejohnson:
I merged the changes from #128 into master and updated `Try.Config.mainGist` to point to my updated fork.

In order to load the modules, it's necessary to copy the contents of `staging/core/.psci_modules/node_modules/` into `client/js/output/`.

@hdgarrood Again this is the bare minimum. The other backends haven't been removed, no READMEs have been updated, and none of the comments from https://github.com/purescript/trypurescript/pull/135#pullrequestreview-363054421 have been implemented.

@natefaubion your changes are great! This will likely allow a lot of flexibility for people wanting to run TryPS locally.

### DIFF
--- a/client/src/Try/Shim.purs
+++ b/client/src/Try/Shim.purs
@@ -1,0 +1,27 @@
+module Try.Shim where
+
+import Data.Tuple (Tuple(..))
+import Foreign.Object (Object)
+import Foreign.Object as Object
+
+type Shim =
+  { url :: String
+  , deps :: Array String
+  }
+
+shims :: Object Shim
+shims = Object.fromFoldable
+  [ Tuple "react"
+      { url: "https://unpkg.com/react@16.6.3/umd/react.development.js"
+      , deps: []
+      }
+  , Tuple "react-dom"
+      { url: "https://unpkg.com/react-dom@16.6.3/umd/react-dom.development.js"
+      , deps: [ "react" ]
+      }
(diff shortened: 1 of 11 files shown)


client/src/Try/Shim.purs
```diff
@@ -0,0 +1,27 @@
...
+      }
+  , Tuple "react-dom"
+      { url: "https://unpkg.com/react-dom@16.6.3/umd/react-dom.development.js"
+      , deps: [ "react" ]
+      }
+  , Tuple "react-dom/server"
+      { url: "https://unpkg.com/react-dom@16.6.3/umd/react-dom-server.browser.development.js"
+      , deps: [ "react" ]
+      }
+  ]
```
natefaubion: These versions probably need to be updated.

hdgarrood:
In what sense have the other backends not been removed? To me, it looks like they have been. The only remnants of the other backends I can find are the files in `client/js` like `react.js` and `mathbox-bundle.js`.

hdgarrood:
This looks great! I'm going to just go ahead and merge it (and update the shim versions).

hdgarrood:
I've just noticed that link clicks no longer work, presumably because of the new iframe. Perhaps we can handle this by listening for the iframe load event and navigating to the new location manually?

hdgarrood:
Oh I see, there's also the contents of `staging/*`

PR #140, opened 2020-03-15
Allow gist navigation

natefaubion:


natefaubion:
Note this looks for `gist.github.com` urls, not ?gist= urls.

client/index.html
```diff
@@ -184,6 +184,12 @@
         clearInterval(sendSources);
       }, { once: true });
 
+      window.addEventListener("message", function(event) {
+        if (event.data && event.data.gistId && /^[0-9a-f]+$/.test(event.data.gistId)) {
+          window.location.search = "gist=" + event.data.gistId;
```
hdgarrood: Do we need to worry about this clobbering other query string parameters? I haven’t worked out what the `session` one does yet.

natefaubion: I just went with this because that's how all the current links work.

natefaubion: I don't think this is supposed to preserve session, which saves your editor to local state.

hdgarrood: Ahh, of course.

PR #143, opened 2020-04-07
Tidying up a bit

hdgarrood:
* Consolidate the client and server READMEs into one file
* Fix local development instructions in README
* Remove some leftover backend stuff
* Update the license year
* Update the default main gist ID to one which uses gist.github.com/:gist-id urls, as expected by #140, so that the links work
* Update "help" link to point to main repo readme, rather than `gh-pages` branch


hdgarrood:
cc @thomashoneyman @gabejohnson if either of you have time and would like to take a look at this?

README.md
```diff
@@ -49,32 +55,49 @@ httpserver 8080 #eg with: alias httpserver='python -m SimpleHTTPServer'
...
+### Compile PureScript code
 
-```
+**POST /compile**
+
+- Request body: PureScript code
+- Response body: Either `{ js: "..." }` or `{ error: "..." }`
+- Status code: 200 (success), even if compilation failed
+
+The response does not use error codes, to make it easier to use the API from another domain using CORS.
```
kritzcreek, line 88: I think in the context of a compilation service a "compiler error" would still be a success either way. So making it easier to do CORS is nice, but I wouldn't need that motivation here.

README.md
```diff
@@ -1,44 +1,50 @@
...
+Most of these features can be controlled not only from the toolbar, but also using the [query parameters](https://en.wikipedia.org/wiki/Query_string):
 
-- Request body: PureScript code
-- Response body: Either `{ js: "..." }` or `{ error: "..." }`
-- Status code: 200 (success)
+- **Load From Gist**: Load PureScript code from Gist id using the `gist` parameter
+    - Example: `gist=37c3c97f47a43f20c548` will load the code from this Gist if the file was named `Main.purs`
 
-The response does not use error codes, to make it easier to use the API from another domain using CORS.
+- **View Mode**: Control the view mode using the `view` parameter
```
thomashoneyman, line 24: We should probably include the default for each of these, ie.

```
Control the view mode using the `view` parameter.

- Options are: `code` | `output` | `both`
- Default is: `both`
- Example: `view=output` will only display the output
```

You could judge this for yourself by loading up the default Try PureScript page, but if you're say crafting a query string based off of this section in the readme it would be useful to see which options you can skip.

README.md
```diff
@@ -1,44 +1,50 @@
...
-### Example
-
-    dist/build/trypurescript/trypurescript 8081 'bower_components/purescript-*/src/**/*.purs'
-    
-# Development
-
-## 1. Client setup
+### 1. Client setup
 
 ```
```
thomashoneyman, line 45:
```suggestion
```sh
```

(Just for slightly nicer highlighting)

README.md
```diff
@@ -49,32 +55,49 @@ httpserver 8080 #eg with: alias httpserver='python -m SimpleHTTPServer'
 open http://localhost:8080
 ```
 
-## 2. Work with local compile server
+### 2. Local compile server setup
 
 ```
```
thomashoneyman, line 60:
```suggestion
```sh
```

thomashoneyman:
I have a few comments (the suggestions can be applied automatically), but they aren't strictly necessary and this otherwise looks good to me.

hdgarrood:
Thanks! I'm going to go ahead and merge this now just to keep things moving but please feel free to comment/open PRs if you have further suggestions.

Issue #144, opened 2020-04-07
use npm for client JS dependencies


hdgarrood:
rather than checking them in

hdgarrood:
Turns out we weren't even using the checked-in JS deps, so #146 just removes them.

hdgarrood:
Resolved by #146

PR #146, opened 2020-04-11
Tidy up client JS dependencies

hdgarrood:
Runtime JS dependencies are obtained as follows:

- dependencies for the main client window (eg jquery, ace) are obtained
  via cdnjs.cloudflare.com, and so we don't need anything beyond than
  the script tags in index.html.
- dependencies needed inside the iframe (eg react, for gists which use
  react-based libraries like purescript-react or purescript-thermite)
  are obtained via the new 'shims' mechanism introduced by #128 / #136.

The files I am removing in this commit do not appear to be in use. I've
also removed non-purescript dependencies from bower.json, as they also
do not appear to be in use.


hdgarrood:
Resolves #144

PR #147, opened 2020-04-11
Use spago for client's PS dependencies

### LINKED ISSUE #145: use spago for client's purescript dependencies

hdgarrood:
Fixes #145


client/spago.dhall
```diff
@@ -0,0 +1,57 @@
...
+  , "st"
+  , "strings"
+  , "tailrec"
+  , "transformers"
+  , "tuples"
+  , "typelevel-prelude"
+  , "unfoldable"
+  , "validation"
+  , "web-html"
+  ]
```
natefaubion, line 54: Are all these dependencies actually needed for the client? Or were these just the package set dependencies for what you could compile against?

hdgarrood: This list of dependencies was generated by `spago init` from the existing `bower.json` file. They might not all be needed, I'm not sure.

Issue #142, opened 2020-04-07
Restore previous backends as examples


hdgarrood:
* thermite
* flare
* mathbox
* slides
* behaviors

thomashoneyman:
With the larger droplet it looks like we should be able to add these backends again, along with the entire PS package set.

hdgarrood:
Oh, just to clarify, I don't want to restore the backends, I just want to include the relevant libraries in the package set and include the relevant JS libraries as shims so that we can have all of the above work without having to reintroduce backends.

thomashoneyman:
Sorry, I misspoke. I also meant reintroducing these libraries, not as “backends”, but just being available in general.

JordanMartinez:
Is this something we're still interested in doing?

hdgarrood:
Provided that the relevant libraries still work in 0.15, I think it'd be nice - especially given that people had to put work in to get these into Try PS in the first place.

PR #150, opened 2020-04-12
Use spago for managing the server package set, attempt 2

hdgarrood:
Like #148, except we keep the same set of dependencies.


staging/spago.dhall
```diff
@@ -0,0 +1,52 @@
...
+    , "math"
+    , "maybe"
+    , "ordered-collections"
+    , "prelude"
+    , "profunctor"
+    , "proxy"
+    , "quickcheck"
+    , "random"
+    , "react"
+    , "react-dom"
```
natefaubion, line 35: Is this mainly for testing the shim stuff? Do we intend to have a process for adding additional libraries from the set. Or perhaps we should just blanket state we are not adding more libraries until we figure out a caching/purging solution for externs?

hdgarrood: I think I originally added these a week or two ago for testing the shim stuff, yeah. I'd like to at least add back libraries which were previously provided by other backends (#142). After that, saying we are not going to add any more libraries until we figure out a caching/purging solution (at which point we can add the whole package set) works for me.

PR #153, opened 2020-04-26
Put everything needed for deploys in deploy bundles

hdgarrood:
The aim of this commit is to allow Travis to create deploy bundles for
both the client and the server on tags so that everything we need for a
deployment is contained in the deploy bundles. There are two separate
bundles: one for the server, which includes the trypurescript binary as
well as nginx configuration and the systemd service file, and one for
the client, which just includes the HTML, CSS, and JS files we serve
from try.purescript.org.

The server bundle already exists; I have only modified it to
additionally include nginx and systemd configuration. The reason for
this is that I would prefer to have as much as reasonably possible of
the production config in the repository, so that we can move towards
more automated deployments, and also to help avoid a situation where
only one person understands enough about the server setup to be able to
administer it.

For the client bundle, I've moved stuff we want to be publicly visible
into a separate public/ directory, mostly because I don't want to
include things like the output/ and node_modules/ directories in the
client bundle (so that it doesn't become too large).

I've also removed the client/CNAME and client/LICENSE files. The CNAME
file is no longer needed: it was only used for GH pages, which I'd like
to try moving away from. The license for the client code is covered by
the LICENSE file at the repo root - LICENSE and client/LICENSE are the
same, except that the client/LICENSE file has the copyright years listed
as 2013-16, so it is redundant.


thomashoneyman:
I see the nginx config in here, but I'm not seeing anything that copies it to a filesystem location or reads from it. Is there a risk of this getting out of sync?

hdgarrood:
I intend to set this up in a similar way to Pursuit, so there will eventually be a script which does copy the nginx and systemd configs into the appropriate locations - I just haven’t gotten around to that yet.

PR #156, opened 2020-05-02
Allow choosing config at build time

hdgarrood:
If `window.location.hostname` is try.purescript.org, then use the
production configuration, otherwise use the development configuration.
This should hopefully do what we want in most cases without requiring
any additional work.


natefaubion:
Maybe we could have two different config modules, and include them as an additional src at build time?

hdgarrood:
Sure, I’ll give that a try.

hdgarrood:
I ended up removing the `bundle` package.json script because `spago bundle-app` performs a build before running the bundling step, and so we'd need to duplicate the arguments across both invocations to get things to work properly, which seems awkward (and compiling without bundling isn't something you want to do very often anyway).

ci/build.sh
```diff
@@ -12,8 +12,8 @@ case $COMPONENT in
   client)
     cd client
     npm install
-    npm run build
-    npm run bundle
+    # Use production config, since we want to use these bundles for deploys
+    npm_config_configpath="config/prod/*.purs" npm run build
```
natefaubion, line 16: I don't know the exact rules of npm env vars. Am I correct in that:
* `npm_config_configpath` is equivalent to overriding the `config.configpath` settings in package.json
* `npm_package_config_configpath` is equivalent to referencing `config.configpath` within the current package.json.

hdgarrood: As I understand it, `npm_config_*` is for configuring the npm client itself, and is only influenced by npmrc files, not anything inside package.json. `npm_package_config_*` is indeed equivalent to referencing `config.configpath` inside `package.json`, unless that configuration option has been overridden by an environment variable, as it has here.

hdgarrood:
Thanks!

PR #159, opened 2020-05-02
Serve the output directory from compile.purescript.org

hdgarrood:


deploy/nginx.conf
```diff
@@ -75,4 +75,8 @@ server {
     location / {
       proxy_pass http://127.0.0.1:8081;
     }
+
+    location /output {
+      alias /var/www/trypurescript/staging/.psci_modules/node_modules;
```
natefaubion, line 80: Is it a concern that this also potentially serves very large externs? Is it possible to glob only js files?

hdgarrood: That's probably not a bad idea. I think it should be possible, I'll play around with it a bit.

hdgarrood: Yep, this is pretty straightforward. There's even an example in the docs which covers pretty much this exact use case: http://nginx.org/en/docs/http/ngx_http_core_module.html#location

PR #162, opened 2020-05-02
Update to PureScript v0.13.6

hdgarrood:


server/Main.hs
```diff
@@ -72,7 +73,8 @@ server externs initEnv port = do
             Right m | P.getModuleName m == P.ModuleName [P.ProperName "Main"] -> do
               (resultMay, ws) <- runLogger' . runExceptT . flip runReaderT P.defaultOptions $ do
                 ((P.Module ss coms moduleName elaborated exps, env), nextVar) <- P.runSupplyT 0 $ do
-                  desugared <- P.desugar externs [P.importPrim m] >>= \case
+                  env <- fmap fst . runWriterT $ foldM P.externsEnv P.primEnv externs
```
natefaubion, line 76: This builds the externs for every compilation, right? Is it possible to build this once and reuse it?

hdgarrood: Ah of course, yes it does.

thomashoneyman:
With this merged, are there any steps we should take to deploy the 0.13.6 version, or is it not quite ready for that?

hdgarrood:
I've deployed this to the DigitalOcean droplet which I intend to be the new home for Try PureScript, and it seems to work! You can try it out by adding the following lines to your `/etc/hosts` file:
```
159.65.220.76 try.purescript.org
159.65.220.76 compile.purescript.org
```
The only thing left to do is to update the gists: the main one still has links to all of the different backends, which no longer exist, so these links won't do anything, and a couple of the other ones have references to library functions which no longer exist, like `id` (which was replaced with `identity`).

I am coming to think that using gists for the examples is a bit awkward, since you can't do pull requests on them and access control on gists is not really a thing either. I think it would be ideal to move them inside the repo, although of course we should continue to support loading code from gists. I'll open a new issue for this.

thomashoneyman:
Ah! I hadn’t realized Try PureScript wasn’t already pointing at the new box. And I think version controlling the stable examples in this repository is a good idea.

hdgarrood:
For now, I've just updated the gists. I think I'll switch the DNS over tomorrow unless anyone notices anything else that's broken :)

hdgarrood:
Oh, we should probably fix #165 first too.

PR #173, opened 2020-05-21
Make the whole package set available

### LINKED ISSUE #154: Support loading/unloading of externs as needed on the server.
In order to support opening up to the wider ecosystem, we need to support someway to load and unload externs on the compilation server. We can't keep the whole set in memory at once (at least not without a much larger budget!).

hdgarrood:
Resolves #154. On my machine locally, with the RTS flags in
`deploy/start`, the resident memory seems to sit fairly stable around
1.89G, which is encouraging. I've tweaked the RTS flags slightly in
order to make better use of the resized droplet, which has 4GB memory
and 2 virtual CPU cores.


natefaubion:
Can we confirm what happens when you try to use a foreign library that we have not shimmed? I honestly have no idea!

hdgarrood:
You get an error in the console, like this:
Screenshot from 2020-05-21 21-01-32

hdgarrood:
I'll add the shim for big-integer. Can people think of others we're missing?

natefaubion:
Do we need to update the react version?

thomashoneyman:
React is on 16.13.1 and the shim here is on 16.13.0, so it’s possible to update but the version is pretty current.

thomashoneyman:
I took a breeze through the package set and the only other libraries that stand out:

- https://github.com/MikeMcl/decimal.js for precise-datetime
- https://github.com/uuidjs/uuid for uuid
- https://github.com/mixer/uuid-validate for uuid

There are some Node, Redis, etc. libraries in here too, but they don't really make sense for Try PureScript anyway.

thomashoneyman:
I've upped the React patch version, and I've added the `uuid` and `decimal.js` packages to the shims. If any others crop up in the future we can always add them later on.

client/src/Try/Shim.purs
```diff
@@ -12,16 +12,28 @@ type Shim =
...
+  , Tuple "big-integer"
+      { url: "https://unpkg.com/big-integer@1.6.48/BigInteger.min.js"
+      , deps: []
+      }
+  , Tuple "decimal.js"
+      { url: "https://unpkg.com/decimal.js@10.2.0/decimal.min.js"
+      , deps: []
+      }
+  , Tuple "uuid"
+      { url: "https://cdnjs.cloudflare.com/ajax/libs/uuid/8.1.0/uuid.min.js"
```
thomashoneyman, line 35: There's nothing on `unpkg` for this package, but it does exist on Cloudflare.

PR #181, opened 2020-05-30
Allow loading examples from GitHub

### LINKED ISSUE #164: Move example gists into the repo
Right now, the code that comes up when you first load Try PureScript is loaded from a gist, whose ID is specified in the configuration (`client/config/dev/Try.Config.purs`). That gist also contains links to other gists.

I think using gists for these examples is a bit awkward, since you can't do pull requests on them, and access control on gists is not really a thing either. I think it would be ideal to move them inside the repo, although of course we should continue to support loading code from gists. Perhaps we could achieve this by including these examples in the deploy bundle and hosting them at try.purescript.org, and then implementing an `example` query string parameter which operates in a similar way to the `gist` one, except that instead of a gist ID to load, it specifies a URL path to a PureScript source file? A query string like `?example=Records` could tell the client to fetch `try.purescript.org/examples/Records/Main.purs` and load it into the editor, for instance.

hdgarrood:
Resolves #164. This commit implements loading module content from GitHub simply by taking a `github` query string parameter, if present, and tacking that string onto the end of `https://raw.githubusercontent.com/` in order to fetch a PureScript source file from somewhere on GitHub. So for example `try.purescript.org/?github=purescript/trypurescript/master/client/examples/Main.purs` would load the file `client/examples/Main.purs` from the `master` branch in this repo.

I'm hoping that by doing it this way, we will make it easier not only for us to maintain examples, but also for people writing documentation for other projects to create and maintain examples that they can link to and collaborate on, since collaborating on GitHub is much easier than collaborating on Gists.

This is a bit rough at the moment, and I've only added the main example in here, which is why I've marked the pull request as draft. I think at least the following needs to happen before this is mergeable:

* Add handling like we have for gists to allow linking to a github file from within the iframe
* Move the rest of the examples into `client/examples`

but before I go ahead and do that I'd like to just check that the approach seems reasonable.

### DIFF
--- a/client/config/dev/Try.Config.purs
+++ b/client/config/dev/Try.Config.purs
@@ -6,5 +6,5 @@ loaderUrl = "js/output"
 compileUrl :: String
 compileUrl = "http://localhost:8081"
 
-mainGist :: String
-mainGist = "7ad2b2eef11ac7dcfd14aa1585dd8f69"
+mainGithubExample :: String
+mainGithubExample = "purescript/trypurescript/master/client/examples/Main.purs"
--- a/client/config/prod/Try.Config.purs
+++ b/client/config/prod/Try.Config.purs
@@ -6,5 +6,5 @@ loaderUrl = "https://compile.purescript.org/output"
 compileUrl :: String
 compileUrl = "https://compile.purescript.org"
 
-mainGist :: String
-mainGist = "7ad2b2eef11ac7dcfd14aa1585dd8f69"
+mainGithubExample :: String
+mainGithubExample = "purescript/trypurescript/master/client/examples/Main.purs"
--- a/client/examples/Main.purs
+++ b/client/examples/Main.purs
@@ -1,0 +1,57 @@
+module Main where
+
+import Prelude
+
+import Effect (Effect)
+import Data.Foldable (fold)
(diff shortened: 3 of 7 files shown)


hdgarrood:
I've verified that this works locally by replacing the `mainGithubExample` in `Try.Config` with this:
```purescript
mainGithubExample :: String
mainGithubExample = "hdgarrood/trypurescript/load-from-github/client/examples/Main.purs"
```

JordanMartinez:
Nice!

natefaubion:
I don't have a solution for this, and maybe I'm just not thinking about it enough, but is it possible to have both content and examples in the same repo, linking to runnable examples from the content and still have both part of a single tag? I know that sounds a little convoluted, but is there a way to use this without always linking to a fixed branch (be it `master`, or `examples`, or whatever)?

natefaubion:
One solution is if you have a publication system, it could potentially rewrite links.

hdgarrood:
I'm not sure I follow. Eventually you'll always have to specify a branch or a tag, right? Are you thinking that it might be awkward to work on guides in separate repositories with this design because you might want to point to a tag that doesn't exist yet?

thomashoneyman:
Closed in #218

PR #182, opened 2020-05-30
Allow any module name

### LINKED ISSUE #174: Allow module names other than Main
I don't think there's any technical reason the module needs to be called Main on the backend. As far as I can tell, the module name in the request just needs to not clash with any module names in the package set.

hdgarrood:
Seems to work. Fixes #174

### DIFF
--- a/server/Main.hs
+++ b/server/Main.hs
@@ -71,7 +71,7 @@ server externs initNamesEnv initEnv port = do
           case CST.parseModuleFromFile "<file>" input >>= CST.resFull of
             Left parseError ->
               return . Left . CompilerErrors . P.toJSONErrors False P.Error $ CST.toMultipleErrors "<file>" parseError
-            Right m | P.getModuleName m == P.ModuleName "Main" -> do
+            Right m -> do
               (resultMay, ws) <- runLogger' . runExceptT . flip runReaderT P.defaultOptions $ do
                 ((P.Module ss coms moduleName elaborated exps, env), nextVar) <- P.runSupplyT 0 $ do
                   desugared <- P.desugar initNamesEnv externs [P.importPrim m] >>= \case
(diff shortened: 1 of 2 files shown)


hdgarrood:
Ah, this behaves oddly if you give the module a name of an existing module in the package set. For example, if I call the current module `Data.Semiring` and try to use `+` within it, then I get a "ReferenceError: $foreign is not defined". Perhaps I should rewrite the module name to `Main` to ensure that it doesn't clash with any module name in the package set?

hdgarrood:
This now seems to work even if I try to redefine an existing module. I haven't been able to think of a way that rewriting the module name like this would be an issue, it seems like a slightly nicer UX for what is essentially the same idea (only modules with a specific module name are allowed).

server/Main.hs
```diff
@@ -71,10 +71,14 @@ server externs initNamesEnv initEnv port = do
           case CST.parseModuleFromFile "<file>" input >>= CST.resFull of
             Left parseError ->
               return . Left . CompilerErrors . P.toJSONErrors False P.Error $ CST.toMultipleErrors "<file>" parseError
-            Right m | P.getModuleName m == P.ModuleName "Main" -> do
+            Right m -> do
+              -- Rewrite the module name to "$Main" in order to ensure that the
+              -- module name doesn't clash with any existing module names in
+              -- the package set.
+              let rewriteModuleName (P.Module ss coms _ decls refs) = P.Module ss coms (P.moduleNameFromString "$Main") decls refs
```
natefaubion, line 78: If someone attempts to play with module export syntax this might be cause problems with:
```
module Foo (module Bar, module Foo) where ...
```

natefaubion: I'll wager this is by far an edge case, however 😆

hdgarrood: Ah, good point. I think re-exporting `main` from elsewhere seems plausible to want to do, so perhaps this isn't the best approach. Maybe we should assemble a set of used module names from the externs files, and then instead of rewriting, just check that the given module name is not included in that set?

Issue #118, opened 2018-03-22
Unable to create a gist


gabejohnson:
When attempting to share, I receive an alert with the message:
```
From try.purescript.org
Failed to create gist
```
This is caused by a failed `POST`:
```
Request URL: https://api.github.com/gists
Request Method: POST
Status Code: 422 Unprocessable Entity
```
The response payload:
```json
{
  "message": "Validation Failed",
  "errors": [
    {
      "resource": "Gist",
      "code": "missing_field",
      "field": "user"
    },
    {
      "resource": "Gist",
      "code": "missing_field",
      "field": "files"
    }
  ],
  "documentation_url": "https://developer.github.com/v3/gists/#create-a-gist"
}
```

soupi:
Thanks for reporting this! Maybe the API for creating an anonymous gist changed. I'll try to take a look at this over the weekend.

soupi:
So [Github removed the option to post anonymous gists](https://blog.github.com/2018-02-18-deprecation-notice-removing-anonymous-gist-creation/).

One possible alternative is to use a different service, such as [lpaste](http://lpaste.net).

joncfoo:
It's not entirely certain what the future of lpaste is - https://www.reddit.com/r/haskell/comments/862k28/looking_for_new_maintainer_of_lpastenet/

Would it be ok to use GitHub OAuth and then create the gist as the editing user?

joncfoo:
The following is a result of the PureScript hack session:

If we decide to go down the oauth route to store gists on behalf of users we'll have to:

1. Bring in the following packages 
- cookie (to encode/decode cookie data from the header)
- clientsession (store user data (tokens) in an encrypted cookie)
- http-client
- http-client-tls
- hoauth2 (used to make oauth2 requests - it's entirely possible not to use this)

0. OAuth application needs to be registed with GitHub in order to receive secret tokens

The flow of code from trypurescript (FE, BE - front-end, back-end) would look like:

** User has a previous session **

The session info is in the secure cookie

1. FE: user clicks share on trypurescript (triggers XHR call to BE)
2. BE: if user has a session and a previous authtoken exists, use it to create a the gist on behalf of the user and report success/failure

** User does not have a previous session **

1. FE: user clicks share on trypurescript (triggers XHR call to BE)
2. BE: requests the FE take the user to github for authentication/authorization (to allow trypurs to create gists - using the `gist` scope)
[...]

PR #187, opened 2020-08-02
Enable gist saving plus other updates

milesfrain:
Live demo at https://try.ps.ai

Will edit this comment with a more complete description of changes, but here's a quick summary:
* Converted more HTML and JS code to PS
* Converted from JQuery to Halogen Hooks
* Using Tailwind CSS
* Enabled gist saving
* Enabled loading files from github repo paths
* Editor state tracked in shareable url, rather than local storage

Address:
* #118 
* #181
* #182 - Simply overwrites the module name client-side. The backend solution proposed in that PR seems more robust though. Example: https://try.ps.ai/?github=JordanMartinez/purescript-cookbook/master/recipes/ButtonsHalogenHooks/src/Main.purs


client/public/index.html
```diff
@@ -1,198 +1,11 @@
...
-    editor.setFontSize(13);
-    editor.setShowPrintMargin(false);
-
-    var session = editor.getSession();
-
-    session.setMode('ace/mode/haskell');
-    session.setOptions({
-      tabSize: 2,
-      useSoftTabs: true
-    });
```
milesfrain, line 114: * showGutter already defaults to true
* original font size is 13, but appears to default to 12. Can't edit this without fixing ace editor library to take a number rather than a string for font size.
* soft tabs enabled by default

client/public/index.html
```diff
@@ -1,198 +1,11 @@
...
-
-  <link rel="icon" type="image/png" href="./img/favicon_clear-16.png" sizes="16x16">
-  <link rel="icon" type="image/png" href="./img/favicon_clear-32.png" sizes="32x32">
-  <link rel="icon" type="image/png" href="./img/favicon_clear-256.png" sizes="256x256">
-
-  <script type="text/javascript" src="//cdnjs.cloudflare.com/ajax/libs/underscore.js/1.8.3/underscore-min.js"></script>
-  <script type="text/javascript" src="//cdnjs.cloudflare.com/ajax/libs/jquery/1.12.4/jquery.js"></script>
-  <script type="text/javascript" src="//cdnjs.cloudflare.com/ajax/libs/ace/1.1.01/ace.js" charset="utf-8"></script>
-  <script type="text/javascript" src="//cdnjs.cloudflare.com/ajax/libs/ace/1.1.01/mode-haskell.js"></script>
-  <script type="text/javascript" src="//cdnjs.cloudflare.com/ajax/libs/ace/1.1.01/theme-dawn.js"></script>
```
milesfrain, line 17: `theme-dawn` does not appear to be used, so I removed it.

client/public/index.html
```diff
@@ -1,198 +1,11 @@
...
-
-  <link rel="stylesheet" type="text/css" href="//fonts.googleapis.com/css?family=Roboto:300,600">
-  <link rel="stylesheet" type="text/css" href="//fonts.googleapis.com/css?family=Roboto+Slab:300,600">
```
milesfrain, line 20: Is this font necessary? I think the default font looks pretty nice in the demo app.

client/package.json
```diff
@@ -1,17 +1,56 @@
 {
-  "name": "trypurescript-client",
-  "private": true,
+  "name": "tps",
```
milesfrain, line 2: Renamed from `trypurescript-client` to make changing configs with these commands a bit less verbose:
```
npm config set tps:configpath "config/dev/*.purs"
npm config set tps:configpath "config/prod/*.purs"
```

client/package.json
```diff
@@ -1,17 +1,56 @@
...
+    "c-user-facing": "# --------- user facing scripts -------",
+
+    "c-gen-css": "# -- Generate tailwind css AND purs css wrapper",
+    "gen-css": "npm run build-css-only && npm run css2purs",
+
+    "c1-lock-css": "# -- Strip away unused css from autogenerated purs files.",
+    "c2-lock-css": "# -- This improves rebuild times, but won't allow adding new css.",
+    "c3-lock-css": "# -- So re-run gen-css before making additional css changes.",
+    "lock-css": "npm run bundle && npm run css-purge && npm run css2purs",
+
```
milesfrain, line 17: It turns out that `npm i` will reformat this file and remove the unnecessary newlines I added for readability. Really terrible that you can't add comments to json files without workaround like this. Would like to migrate to something better if it exists.

client/src/Try/Component.purs
```diff
@@ -0,0 +1,542 @@
...
+            obj <- liftAff $ runLoader loader $ JS js
+            -- Compress source file contents.
+            -- This avoids non-trivial HTML + JSON string escaping
+            let
+              objcomp = map (unwrap >>> compressToEncodedURIComponent) obj
+            -- Convert object to JSON string
+            let
+              jsonStr = stringify $ encodeJson objcomp
+            -- Pass JS files to execute in iframe
+            putOutput $ App jsonStr $ JS js
```
milesfrain, line 164: The JS files are compressed with lz-string because I couldn't figure out how to escape the contents correctly to pass to the iframe. It would be nice to escape these a bit more efficiently, although I don't think this compression/decompression step is a major performance issue.

client/src/Try/Common.purs
```diff
@@ -0,0 +1,61 @@
...
+
+import Prelude
+
+{-
+Common values and newtype wrappers
+-}
+--
+-- Page to launch on startup and when clicking home button
+homeRoute :: String
+homeRoute = "/?github=milesfrain/tps/demo/examples/Home.purs"
```
milesfrain, line 11: Considering changing these paths to expect the `blob` part so they're easier for users to copy while browsing a repo. Then it just needs to be stripped out from the `raw` url for loading.
```
https://github.com               /milesfrain/tps/blob/demo/examples/Home.purs
https://raw.githubusercontent.com/milesfrain/tps/     demo/examples/Home.purs
```

client/src/Try/Loader.purs
```diff
@@ -92,32 +111,51 @@ makeLoader rootPath = Loader (go Object.empty <<< parseDeps "<file>")
...
+                where
+                src = JS $ body <> "\n//# sourceURL=" <> path'
+          -- No path means dependency is a shim
+          Nothing -> case Object.lookup name shims of
+            Just shim -> do
+              res <- AX.get AXRF.string shim.url
+              case res of
+                Left err -> pure { name, path, deps: [], src }
+                  where
+                  src = throwJSError $ "Could not get shim " <> name <> " at " <> shim.url <> ", " <> AX.printError err
```
milesfrain, line 143: This should point to a guide on adding new shims

client/src/Try/Loader.purs
```diff
@@ -65,23 +68,39 @@ parseDeps current = Array.mapMaybe go <<< String.split (Pattern "\n") <<< unwrap
...
+  -- Run loader to collect all dependencies for compiled code
+  obj <- k js
+  -- Return dependencies along with compiled code
+  pure $ Object.insert "<file>" js obj
+
+makeLoader :: Loader
+makeLoader = Loader (go Object.empty <<< parseDeps "<file>")
   where
   moduleCache :: Ref (Object Module)
   moduleCache = unsafePerformEffect (Ref.new Object.empty)
```
milesfrain, line 106: I was hoping there would be a nice way to do this with [memoize](https://pursuit.purescript.org/packages/purescript-memoize/5.0.0), but I couldn't figure it out. It doesn't seem like memoize can be used to bypass repeating effects (such as http requests in this case).

hdgarrood:
This is very difficult to review as-is because of its size. Can you please break this up into smaller PRs which address individual points? Also can you please back out the changes which already exist in other PRs? It's better to just get those changes in one at a time.

thomashoneyman:
Thanks for doing this, @milesfrain! I'd love to review this and I'm happy to see some improvements coming in. As @hdgarrood said, it's a little difficult to review a PR of this size, and it would be much easier to review and faster to merge PRs for individual points if it's possible to break this up.

milesfrain:
I agree that a giant wave of changes is not ideal.

I can back out these feature additions:
* Gist saving
* File loading
* Any module name

But splitting-apart the remaining changes would involve a lot of throw-away effort to reach a working state between each PR and would be more code to review overall.

The only code that ended up being reused across the rewrite is:
* `src/Try/Loader.purs`
* `src/Try/Types.purs`
* `src/Try/Shim.purs`
* `config/*/Try.Config.purs`
* `src/Try/API.*` is now `src/Try/Compile.purs`
* `src/Try/Gist.*` is now `src/Try/Request.purs`
* `public/js/frame.js` is now `public/frame-load.js`

The code may be easiest to review by considering it as a full rewrite and exploring it as a fresh new project locally, rather than as an incremental change in the diff viewer. It's 1400 lines of commented PS code.

No rush on getting this merged, since these features are available in the live demo, and I'm happy to use that in the meantime. I also don't want to take attention away from other tasks. I'll still do my best to help with the merge effort - starting with making another PR containing fewer new features.

hdgarrood:
I would rather have more code to review overall if that allows making changes incrementally; I don't want to do a full rewrite. Most of the individual changes you've listed do make the most sense to do incrementally, in my mind. Also, I don't want to make changes as far-reaching as moving from jquery to halogen hooks, or moving to tailwind css, or changing the iframe strategy, without a proper justification. I think each of these things really needs its own issue to make the case for why it's a good idea.

milesfrain:
I split this up into six PRs: #190 though #195.

#192 is still pretty significant.

If this needs to be broken-up further, we should first agree on a commit sequence. This is one option:
1. webpack
1. iframe
1. jquery -> halogen
    - and saving state in URL
1. tailwind css

But there's a pretty significant time cost to doing this, so I don't want to dive into that unless it's the only way to move forward with incorporating these changes.

PR #192, opened 2020-08-04
Refresh with Halogen, Tailwind, Webpack

milesfrain:
This PR aims to:
* Trade JS and HTML code for PS
* Improve type safety
  * The original version uses JQuery, which passes around lots of strings that aren't checked during compilation. This new version uses Halogen, which is a more type safe framework.
  * Tailwind improves type safety for CSS. It's also a great way to give PS code more control over styling.
* Improve the developer workflow
  * Webpack dev server allows for fast incremental rebuilds and automatic page refresh.
* Simplify iframe strategy
  * The original version involved bidirectional communication between iframe and parent, along with synchronization code (timeouts, retries, etc). This seems unnecessary. The new architecture allows the parent to simply dictate iframe content upon creation.

The core changes are in the last commit of this PR. The other commits are these PRs:
#190 
#191

These PRs track additional features that may be added after this PR:
#193
#194
#195


thomashoneyman:
This was mostly done in #215, with iframe changes done in #275.

PR #194, opened 2020-08-04
Enable loading GitHub repo paths

milesfrain:
A version of #181 applied to the Halogen rewrite (#192).

The last commit of this PR contains the relevant changes.

Intended to be merged after #193 

Demo available at [try.ps.ai](https://try.ps.ai)

PR #195, opened 2020-08-04
Automatically overwrite modules as Main

milesfrain:
This PR is an alternate version of #182 applied to the Halogen rewrite (#192). This version simply overwrites the module name client-side. The backend solution proposed in #182 seems more robust, so happy to go with that solution instead..

The last commit of this PR contains the relevant changes. The diff will be clearer after #194 is merged.

Example: https://try.ps.ai/?github=JordanMartinez/purescript-cookbook/master/recipes/ButtonsHalogenHooks/src/Main.purs


hdgarrood:
yeah, I'd prefer to do this on the backend.

thomashoneyman:
Done in #182

Issue #183, opened 2020-06-26
Are development instructions still accurate?


milesfrain:
Following these steps: https://github.com/purescript/trypurescript/blob/master/README.md#development

Compilation seems to work, but then nothing is displayed in right pane.

Seems to be due to requesting files that don't exist.

From browser:
```
Request URL: http://localhost:8080/js/output/Data.Foldable/index.js
Request Method: GET
Remote Address: 127.0.0.1:8080
Status Code: 404
```
From `httpserver` logs:
```
127.0.0.1 - - [26/Jun/2020 14:03:21] code 404, message File not found
127.0.0.1 - - [26/Jun/2020 14:03:21] "GET /js/output/Data.Foldable/index.js HTTP/1.1" 404 -
127.0.0.1 - - [26/Jun/2020 14:03:21] code 404, message File not found
127.0.0.1 - - [26/Jun/2020 14:03:21] "GET /js/output/Data.Functor/index.js HTTP/1.1" 404 -
127.0.0.1 - - [26/Jun/2020 14:03:21] code 404, message File not found
127.0.0.1 - - [26/Jun/2020 14:03:21] "GET /js/output/TryPureScript/index.js HTTP/1.1" 404 -
```

There's no `output` directory where it's looking:
```
> ls client/public/js 
frame.js  index.js
```

milesfrain:
Here's one workaround:
```
pushd staging
spago build
popd

pushd client/public/js
ln -s ../../../staging/output output
popd
```

Although I notice that the production instance is grabbing these files via the compiler server, e.g:
https://compile.purescript.org/output/Control.Bind/index.js

So I also tried this edit in `client/config/dev/Try.Config.purs`:
```diff
-loaderUrl = "js/output"
+loaderUrl = "http://localhost:8081/output"
```

But that didn't work out.

hdgarrood:
I think doing a symbolic link is the way to go. As for changing Config.purs: it doesn't really matter whether those files are being grabbed from the compile API host or the web frontend host. If you want to grab them from the compile host then you'll need nginx or something in front of the Haskell backend to do that, which isn't worth the hassle in development, I would say.

thomashoneyman:
@hdgarrood I went to hack on Try PureScript tonight, but following the development instructions:

```sh
git clone git@github.com:purescript/trypurescript.git
cd trypurescript

stack build

# Install PureScript dependencies
cd staging
spago install

stack exec trypurescript 8081 $(spago sources)
[...]

PR #196, opened 2020-08-22
Add missing spago build step

milesfrain:
An `output` directory is necessary and `spago build` creates this.
Note that `spago build` runs `spago install` first.

This doesn't fully fix development instructions, so leaving this as a draft.

Related to #183.

### DIFF
--- a/README.md
+++ b/README.md
@@ -76,7 +76,7 @@ stack build
 
 # Install PureScript dependencies
 cd staging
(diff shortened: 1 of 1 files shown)


JordanMartinez:
I think this can be closed as the issue is due to source globs as discovered in #220 and fixed in #221. Starting the server correctly will build the files.

PR #198, opened 2020-08-22
One-way iframe communication

milesfrain:
This PR "simplifies" iframe management by eliminating some JS code, and letting PS code setup the iframe directly. The previous version required some two-way message passing, along with retries and timeouts. More critiques of the previous version can be found in https://github.com/purescript/trypurescript/pull/192#issuecomment-678673743.

There are still some non-ideal aspects of this new strategy, such as adding another layer of encoding.

### DIFF
--- a/client/src/Main.purs
+++ b/client/src/Main.purs
@@ -204,8 +204,8 @@ execute js modules = do
   -- Set up an iframe and use it to execute the provided JavaScript code.
   column2 <- JQuery.select "#column2"
   iframe <- JQuery.create "<iframe>"
-  -- allow-top-navigation lets us click on links to other examples
-  JQuery.setAttr "sandbox" "allow-scripts allow-top-navigation" iframe
+  -- allow-top-navigation-by-user-activation lets us click on links to other examples
+  JQuery.setAttr "sandbox" "allow-scripts allow-top-navigation-by-user-activation" iframe
   JQuery.setAttr "src" "frame-error.html" iframe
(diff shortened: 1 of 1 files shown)


hdgarrood:
@milesfrain could you please fix this on try.ps.ai by removing allow-top-navigation and allow-top-navigation-on-user-action there? I see that try.ps.ai is getting a fair bit of use on FP Slack already and it's quite easy to exploit this there right now.

milesfrain:
Eliminating `top-navigation` from the iframe will require a longer dev session, since the current architecture relies on a declarative way of setting up the iframe. I should have time for that after next Tuesday.

For now, I made a quick fix to switch to the less egregious `allow-top-navigation-by-user-activation`.

Another easy (partial and temporary) solution is to add a toggle (disabled by default) to enable links within the iframe.

Taking a step back, I wonder if we even need to support links within the iframe at all? We could simplify the code a lot if we either just relied on a separate page that lists examples (similar to how the cookbook lists recipes).

We could further polish this up by making the examples links more accessible, like what's done for TypeScript and ReasonML:
* https://reasonml.github.io/en/try
* https://www.typescriptlang.org/play

hdgarrood:
I'm not sure I'd want to remove the ability to link from the iframe entirely, as other people might be using it elsewhere, and I don't see a pressing need to simplify this code. However, having a list of examples as part of the actual Try PureScript UI rather than relying on links we render inside the iframe does sound good to me; that way it will probably be easier to navigate around them too.

PR #206, opened 2020-12-12
Remove unused pragmas, imports, and definitions

thomashoneyman:
I finally got Try PureScript building on my machine and took some time to review the server code. While there, I noticed that there are a number of unused language pragmas, imports, and definitions in the file. This PR simply removes the dead code from the file (and removes one unnecessary dollar sign).

### DIFF
--- a/server/Main.hs
+++ b/server/Main.hs
@@ -1,17 +1,13 @@
 {-# LANGUAGE DataKinds #-}
-{-# LANGUAGE DeriveAnyClass #-}
 {-# LANGUAGE DeriveGeneric #-}
 {-# LANGUAGE LambdaCase #-}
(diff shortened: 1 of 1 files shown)


server/Main.hs
```diff
@@ -123,10 +112,7 @@ server externs initNamesEnv initEnv port = do
 
 getOpts :: Int -> Scotty.Options
 getOpts port = def
-  { settings =
-      Warp.setHost "127.0.0.1"
```
hdgarrood, line 127: Can you revert this change please? Each setting we’re overriding ought to have its own line here I think. I understand that the $ is unnecessary here but to me having each setting on its own line is more important.

hdgarrood (approved):

Issue #199, opened 2020-08-23
Replace jQuery with Halogen


thomashoneyman:
Try PureScript currently relies on the [`jquery`](https://pursuit.purescript.org/packages/purescript-jquery/5.0.0/docs/JQuery) library and some extras defined in [`JQuery.Extras`](https://github.com/purescript/trypurescript/blob/91276af46a16234702c7d80f81e438f4d130b82e/client/src/JQuery/Extras.purs) in this application. It uses these bindings largely in the [client/Main.purs](https://github.com/purescript/trypurescript/blob/91276af46a16234702c7d80f81e438f4d130b82e/client/src/Main.purs) file which manages the Try PureScript window and iframe.

There are a few drawbacks to using jQuery:

* It relies on an imperative approach to updating the editor and application
* It relies on a set of string selectors without much guidance via types or documentation about what selectors to use to control what behavior
* (It is my belief that) fewer people in the PureScript community are familiar with using jQuery to write an application like this, versus more common PureScript SPA solutions like React Basic or Halogen.

There are some advantages of jQuery, too:

* More people know jQuery in general, and so it may be the best option as far as encouraging new contributors
* jQuery is more stable and changes less frequently than Halogen or React

I think that using jQuery is a net negative for new folks trying to understand Try PureScript and contribute new features to it. Ultimately it’ll be PureScript developers contributing and my sense is that within the community frameworks like React and Halogen are considerably more popular — and while jQuery may be much more widely used and understood than those in general, within the community there are a lot of folks who weren’t web developers before PureScript (like myself) and are not as familiar with things like jQuery.

I also think that the use of jQuery lends to a style of programming that makes this app harder to grow and maintain over time — specifically the imperative style and having to know what a selector means rather than the more PureScript-y declarative nature of React or Halogen.

For these reasons I think we should explore rewriting Try PureScript in either React Basic or Halogen. This rewrite would add no new features and faithfully preserve how Try PureScript currently works, except in the React style or the Halogen style.

(Also — I think it makes sense to switch `ContT` for `Aff` at the same time, for ease of development. We’d already agreed that this is worth doing in https://github.com/purescript/trypurescript/issues/179 and it will probably be easier to make both changes together rather than change one at a time.)

[...]

PR #208, opened 2020-12-15
Migrate from ContT to Aff

thomashoneyman:
This PR addresses the first part of #199 by switching from `ContT` and the FFI for requests to `Aff` and Affjax for requests throughout the application. I tried to make the minimum changes required to switch.

The `Main.purs` and `Loader.purs` files have largely mechanical changes, and you can verify at a glance that the changes come down to a few `liftEffect`s and switching `runContT` for Aff binds and the occasional `launchAff_`.

However, the `API.purs` and `Gist.purs` have larger changes because I've switched to use Affjax instead of relying on the FFI for these requests. In those files, you can see that the parameters used for the GET or POST requests have been copied almost exactly into the equivalent Affjax code. I'm leaving a few review comments below to clear up some points that may be confusing at a glance.

I've tested Try PureScript locally using these changes and it continues to work as before. However, I haven't tested saving a gist -- that already doesn't work on Try PureScript today, and I don't think my changes make it any more broken than it was before. A subsequent PR as part of the changes in #199 ought to fix the gist saving issue, as @milesfrain has already demonstrated on try.ps.ai.


client/src/Main.purs
```diff
@@ -132,31 +133,32 @@ compile = do
...
-                             position <#> \(ErrorPosition pos) ->
-                               { row: pos.startLine - 1
-                               , column: pos.startColumn - 1
-                               , type: "warning"
-                               , text: message
-                               }
-                       runEffectFn1 setAnnotations (mapMaybe toAnnotation warnings_)
-                     for_ sources (execute (JS js))
-          Right (CompileFailed (FailedResult { error })) -> do
+            showJs <- liftEffect isShowJsChecked
```
thomashoneyman, line 144: These changes are largely indentation related; the only real difference is the introduction of `liftEffect` and switching `runContT` for binds in `Aff`.

client/src/Try/API.purs
```diff
@@ -123,31 +125,23 @@ instance decodeCompileResult :: Decode CompileResult where
...
-       (Either (NonEmptyList ForeignError) CompileResult)
-compile endpoint code = ExceptT (ContT \k -> runEffectFn4 compile_ endpoint code (mkEffectFn1 (k <<< Right <<< runExcept <<< decode)) (mkEffectFn1 (k <<< Left)))
+get :: URL -> ExceptT String Aff String
+get url = ExceptT $ AX.get AXRF.string url >>= case _ of
+  Left e ->
+    pure $ Left $ printError e
+  Right { status } | status >= StatusCode 400 ->
+    pure $ Left $ "Received error status code: " <> show status
+  Right { body } ->
+    pure $ Right body
```
thomashoneyman, line 135: I believe the second check (which guards against status codes over 400) is technically unnecessary, but I couldn't verify that at a quick breeze through the Affjax documentation so I've included it here. I can remove these checks if they're redundant, though.

This doesn't really need to be in `ExceptT`, but doing so allows me to minimize the diff everywhere else.

client/src/Try/API.purs
```diff
@@ -123,31 +125,23 @@ instance decodeCompileResult :: Decode CompileResult where
...
+
+-- | POST the specified code to the Try PureScript API, and wait for a response.
+compile :: String -> String -> ExceptT String Aff (Either (NonEmptyList ForeignError) CompileResult)
+compile endpoint code = ExceptT $ AX.post AXRF.json (endpoint <> "/compile") (Just requestBody) >>= case _ of
+  Left e ->
+    pure $ Left $ printError e
+  Right { status } | status >= StatusCode 400 ->
+    pure $ Left $ "Received error status code: " <> show status
+  Right { body } ->
+    pure $ Right $ runExcept (decode (unsafeToForeign body))
```
thomashoneyman, line 145: This `unsafeToForeign` is for parity with the previous FFI code, which specified a JSON return type (in jQuery) and then decodes it as `Foreign` (from PureScript).

client/src/Try/Gist.js
```diff
@@ -1,48 +1,10 @@
...
-      fail(err.statusText);
-    });
-  } else {
-    fail("Gist does not contain a file named " + filename);
+exports.rawUrl_ = function (gistInfo_, filename) {
+  try {
+    var gistInfo = JSON.parse(gistInfo_);
+    return gistInfo.files[filename].raw_url;
+  } catch (e) {
+    return null;
```
thomashoneyman, line 8: I opted to keep this code in the FFI rather than implement the decoding in PureScript, once again to keep the diff minimal. In the future, though, it would be nice to switch to something like `codec-argonaut` instead.

client/src/Try/Gist.purs
```diff
@@ -5,45 +5,78 @@ module Try.Gist
...
+  Left e ->
+    pure $ Left $ "Unable to load Gist metadata: \n" <> printError e
+  Right { status } | status >= StatusCode 400 ->
+    pure $ Left $ "Received error status code: " <> show status
+  Right { body } ->
+    pure $ body # caseJsonObject (Left $ "Expected object in uploadGist, received: " <> stringify body) \obj ->
+      case Object.lookup "id" obj of
+        Just v | Just v' <- toString v -> Right v'
+        Nothing -> Left "No id key found."
+        _ -> Left "Key id was not a string."
```
thomashoneyman, line 55: This seemed like the simplest way to pull out the `id` key from an object in the response without resorting to the FFI and without writing a new decoder. Like the `rawUrl_`  FFI code I think ideally we'd move to something like `codec-argonaut` in the future and replace this with a proper representation of the response from the API.

thomashoneyman (commented):
I've left some additional clarifying comments.

thomashoneyman:
One last thing: I needed to symlink the `staging/output` directory into `public/js/output` in order for Try PureScript to pick up dependencies. Simply running `spago install` and then `stack exec trypurescript 8081 $(spago sources)` wasn't enough (as per the dev instructions in the README). Just a FYI for anyone testing this out locally.

client/src/Try/Gist.js
```diff
@@ -1,48 +1,9 @@
...
-  $.ajax({
-    url: 'https://api.github.com/gists/' + id,
-    dataType: 'json'
-  }).done(done).fail(function(err) {
-    fail("Unable to load Gist metadata");
-  });
-}
-
-exports.tryLoadFileFromGist_ = function(gistInfo, filename, done, fail) {
+exports.rawUrl_ = function (gistInfo, filename) {
```
thomashoneyman, line 3: This small function, `rawUrl_`, is copied from the beginning of `tryLoadFileFromGist_`:

```js
exports.tryLoadFileFromGist_ = function(gistInfo, filename, done, fail) {
  if (gistInfo.files && gistInfo.files.hasOwnProperty(filename)) {
    var url = gistInfo.files[filename].raw_url;
```

I opted to keep this code in the FFI rather than implement the decoding in PureScript, once again to keep the diff minimal. In the future, though, it would be nice to switch to something like `codec-argonaut` instead.

milesfrain (approved):

JordanMartinez (approved):

client/src/Main.purs
```diff
@@ -214,33 +215,32 @@ setupEditor { code } = do
...
-  -> Effect Unit
-loadFromGist id_ k = do
-  runContT (runExceptT (getGistById id_ >>= \gi -> tryLoadFileFromGist gi "Main.purs")) $
-    case _ of
-      Left err -> do
-        window >>= alert err
-        k { code: "" }
-      Right code -> k { code }
+  -> ({ code :: String } -> Aff Unit)
+  -> Aff Unit
```
hdgarrood, line 226: Did you consider changing this function to have the type `String -> Aff { code :: String }` or perhaps even `String -> Aff String`? I think callback ought not to be necessary any more now that we're working in Aff; the benefit of Aff is that it lets us not bother with callbacks, so I think we might as well take advantage of that?

client/src/Main.purs
```diff
@@ -214,33 +215,32 @@ setupEditor { code } = do
...
+    Left err -> do
+      liftEffect $ window >>= alert err
+      k { code: "" }
+    Right code -> k { code }
 
 withSession
   :: String
-  -> ({ code :: String } -> Effect Unit)
-  -> Effect Unit
+  -> ({ code :: String } -> Aff Unit)
```
hdgarrood, line 236: Same here re getting rid of the callback

client/src/Try/Gist.purs
```diff
@@ -5,45 +5,78 @@ module Try.Gist
...
-               String
-               (EffectFn1 String Unit)
-               (EffectFn1 String Unit)
-               Unit
-
-tryLoadFileFromGist :: GistInfo -> String -> ExceptT String (ContT Unit Effect) String
-tryLoadFileFromGist gi filename = ExceptT (ContT \k -> runEffectFn4 tryLoadFileFromGist_ gi filename (mkEffectFn1 (k <<< Right)) (mkEffectFn1 (k <<< Left)))
+type FileName = String
+
+type RawURL = String
```
hdgarrood, line 31: I think we should avoid type synonyms like these: they give a false sense of security and I think they require you to keep more in your head while reading the code. In fact we came quite close to making type synonyms like these a warning in the compiler once: https://github.com/purescript/purescript/pull/2932.

If it's important enough to distinguish these types to stop you from putting one in where you meant the other, or for documentation purposes, then at least I think they should be newtypes. Otherwise I think we should just use `String`.

client/src/Try/API.purs
hdgarrood: Can you check and see what happens if you get a 404 status code? I haven't been able to verify from a quick glance through the documentation either but I actually suspect it's not redundant. Either way, I think we should know the answer before merging.

client/src/Main.purs
thomashoneyman: I can go ahead and update these so they aren’t using callbacks anymore.

client/src/Try/API.purs
thomashoneyman: I’ll verify this check is necessary. Do you have any opinion on the code as-is if the check does turn out to be necessary?

client/src/Try/Gist.purs
thomashoneyman: I think this could easily go either way. I’ll think a bit more on this when I come back to it, but I’ll probably revert to String rather than add newtypes. I agree about the false sense of security.

client/src/Try/API.purs
hdgarrood: The code seems fine to me as-is if the check is necessary. If it's not necessary, I think we should remove it.

thomashoneyman: @hdgarrood Yep, it turns out this check is necessary.

If the request is made with a malformed URL:
Screen Shot 2020-12-15 at 6 14 49 PM

If the request returns a 404:
Screen Shot 2020-12-15 at 6 17 01 PM

thomashoneyman:
@hdgarrood I've addressed your feedback

hdgarrood (approved):
Nice :tada:

PR #209, opened 2021-01-23
Update for PureScript 0.14

thomashoneyman:
**Description of changes:**

This PR updates Try PureScript for compatibility with PureScript 0.14.x. In this case I'm using the latest commit for 0.14.0-rc5 in Stack via GitHub. The changes are straightforward, but I'm now throwing away warnings (which weren't previously present) so that's worth a look to see if we can make better use of them.

I haven't been able to thoroughly test this yet, but if I can load up the [prepare-0.14 package set](https://github.com/purescript/package-sets/tree/prepare-0.14) locally for testing purposes then I will report those results back.

Unlike Pursuit, we won't want to deploy this until after 0.14 is out and we can rely on a new package set. But I still wanted to get this ready for release so we're ready to deploy as soon as that package set is together.

Thanks also to @colinwahl for working on this with me.


server/Main.hs
```diff
@@ -102,7 +107,7 @@ server externs initNamesEnv initEnv port = do
               search = fst . TS.typeSearch (Just []) initEnv (P.emptyCheckState initEnv)
               results = nubBy ((==) `on` fst) $ do
                 elab <- elabs
-                let strictMatches = search (replaceTypeVariablesAndDesugar (\nm s -> P.Skolem P.NullSourceAnn nm s (P.SkolemScope 0)) elab)
+                let strictMatches = search (replaceTypeVariablesAndDesugar (\nm s -> P.Skolem P.NullSourceAnn nm Nothing s (P.SkolemScope 0)) elab)
```
thomashoneyman, line 110: Skolems are now tagged with their kinds as per [the PolyKinds PR](https://github.com/purescript/purescript/pull/3779), so there's an additional argument:

https://github.com/purescript/purescript/blob/62c5862430f91a9249046487bd80bd058b3df915/lib/purescript-ast/src/Language/PureScript/Types.hs#L73

If setting this to `Nothing` is the wrong approach, I'd appreciate a note on where I might find this information.

server/Main.hs
```diff
@@ -154,7 +159,7 @@ tryParseType = hush . fmap (CST.convertType "<file>") . runParser CST.parseTypeP
 
     runParser :: CST.Parser a -> Text -> Either String a
     runParser p =
-      first (CST.prettyPrintError . NE.head)
+      bimap (CST.prettyPrintError . NE.head) snd
```
thomashoneyman, line 162: As above, we now have a tuple with `[CST.ParserWarning]`; I opted to throw these warnings away, following the lead of `Language.PureScript.Interactive.Parser`:

https://github.com/purescript/purescript/blob/7ecc42669c69682996f2196ba2eef6c4ca827348/src/Language/PureScript/Interactive/Parser.hs#L66-L71

server/Main.hs
```diff
@@ -53,31 +53,36 @@ instance A.ToJSON Error
...
-                Right js -> (return . Right) (P.toJSONErrors False P.Error ws, js)
-            Right _ ->
-              (return . Left . OtherError) "The name of the main module should be Main."
+              return $ toCompilerErrors parseError
+            Right partialResult -> case CST.resFull partialResult of
+              (_, Left parseError) ->
+                return $ toCompilerErrors parseError
+              (warnings, Right m) | P.getModuleName m == P.ModuleName "Main" -> do
+                (resultMay, ws) <- runLogger' . runExceptT . flip runReaderT P.defaultOptions $ do
+                  ((P.Module ss coms moduleName elaborated exps, env), nextVar) <- P.runSupplyT 0 $ do
```
hdgarrood, line 71: This part of the code is basically an inlined version of `Language.PureScript.Make.rebuildModule`, but `rebuildModule` [has changed inside the compiler](https://github.com/purescript/purescript/compare/v0.13.8...v0.14.0-rc5#diff-fae31b1b9fa5fe55802867766a0c2f98487abbaab050fe24fc40edb162e33d9e) since this was first written so I think this is now not quite right. I think previously this was mostly-inlined from the compiler because using rebuildModule directly would require coming up with a `MakeActions` record, and in particular we'd need to provide a `codegen` field which captures the JS output in a mutable variable rather than writing it to disk. I think previously that probably didn't seem worth the effort, but now that `rebuildModule` has become more complex, it's probably better to take this route.

The `Make` monad in Language.PureScript.Make should be an appropriate choice for the monad type variable, and you can use `runMake` like we already did elsewhere in this file to get back down to IO. Let me know if you'd like to pair on this.

server/Main.hs
```diff
@@ -53,31 +53,36 @@ instance A.ToJSON Error
...
-              case resultMay of
-                Left errs -> (return . Left . CompilerErrors . P.toJSONErrors False P.Error) errs
-                Right js -> (return . Right) (P.toJSONErrors False P.Error ws, js)
-            Right _ ->
-              (return . Left . OtherError) "The name of the main module should be Main."
+              return $ toCompilerErrors parseError
+            Right partialResult -> case CST.resFull partialResult of
+              (_, Left parseError) ->
+                return $ toCompilerErrors parseError
+              (warnings, Right m) | P.getModuleName m == P.ModuleName "Main" -> do
```
hdgarrood, line 69: It might help to call the warnings variable `parserWarnings`? Just because there's quite a lot that goes on in between it being bound here and it being used at the end of this expression.

server/Main.hs
```diff
@@ -53,31 +53,36 @@ instance A.ToJSON Error
 
 server :: [P.ExternsFile] -> P.Env -> P.Environment -> Int -> IO ()
 server externs initNamesEnv initEnv port = do
+  let toCompilerErrors :: NE.NonEmpty CST.ParserError -> Either Error a
```
hdgarrood, line 56: Even if we always want to feed the result of this to `Left`, I think it would be preferable to have this function be `fromParserErrors :: NE.NonEmpty CST.ParserError -> Error`. That way, this function has a clearer responsibility, and it's also clearer at the call site that we're always producing a `Left`.

thomashoneyman: I've floated it to the top level and named it as you suggest. Thanks!

server/Main.hs
thomashoneyman: I'm going to take a stab at this right now, but if I'm stumped then I'd appreciate pairing.

thomashoneyman: There's quite a lot going on in `MakeActions`! I think it would help me to pair on this, but I've stubbed out an implementation using `IORef` for the time being so there's a place to start.

hdgarrood (approved):

PR #211, opened 2021-03-19
Add fixtures for API responses

thomashoneyman:
The PureScript 0.14 release helped drive the point home, for me, that core libraries and infrastructure ought to use core libraries (or contrib, or node, etc.) wherever possible.

As part of the updates originally outlined in #199 I thought we could switch to use `argonaut` rather than `foreign-generic` for the API in Try PureScript. We already depend on `argonaut` and are receiving its `Json` type via Affjax, so it feels like a more natural fit than `foreign-generic` in addition to being a contrib library.

This PR takes a first step by adding API response fixtures and tests for the existing `foreign-generic` implementation. This helps ensure that switching to a new library will still decode correctly without having to test manually.

### DIFF
--- a/client/package.json
+++ b/client/package.json
@@ -6,9 +6,9 @@
   },
   "scripts": {
     "clean": "rimraf output",
-    "test": "spago test --path ./config/dev/Try.Config.purs",
-    "build:dev": "spago build --path ./config/dev/Try.Config.purs",
-    "build:prod": "spago bundle-app --path $npm_package_config_configpath --purs-args '--censor-lib --strict' --to public/js/index.js"
+    "test": "spago test --path $npm_package_config_configpath",
+    "build": "spago build --path $npm_package_config_configpath",
+    "bundle": "spago bundle-app --path $npm_package_config_configpath --purs-args '--censor-lib --strict' --to public/js/index.js"
   },
   "devDependencies": {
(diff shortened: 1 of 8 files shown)


client/package.json
```diff
@@ -6,7 +6,9 @@
   },
   "scripts": {
     "clean": "rimraf output",
-    "build": "spago bundle-app --path $npm_package_config_configpath --purs-args '--censor-lib --strict' --to public/js/index.js"
+    "test": "spago test --path $npm_package_config_configpath",
+    "build": "spago build --path $npm_package_config_configpath",
```
thomashoneyman, line 10: Could we switch this environment variable to something a little shorter? Perhaps `TRY_CONFIG`?

hdgarrood: The reason we have the `npm_package_config_` prefix is so that it can pick up the values from the `config` key in package.json by default if they haven't been explicitly overridden (e.g. like we do in CI: https://github.com/purescript/trypurescript/blob/52c81fc168d2e39f701ed2aaa61694147f6024bc/ci/build.sh#L16). We have to include the `npm_package_config_` prefix for this to work. If there's another way to make `npm run build` Just Work in development and for the production config to be used in CI I'm happy to consider that.

hdgarrood (approved):

PR #212, opened 2021-03-19
Migrate to argonaut-codecs from foreign-generic

thomashoneyman:
This PR continues #211 by migrating from `foreign-generic` to `argonaut-codecs` for JSON encoding and decoding. It also updates the tests and ensures they're exercised in CI.


client/spago.dhall
```diff
@@ -6,25 +6,19 @@ You can edit this file as you like.
 , dependencies =
   [ "aff"
   , "affjax"
+  , "argonaut-codecs"
```
thomashoneyman, line 9: I noticed that a number of dependencies in this file appear to be outdated (we don't use `exists`, `foreign-generic`, `foreign`, `free`, `quickcheck`, and so on anywhere in the code). I've removed them and everything continues to build just fine, but I can replace any of these that we want explicitly.

client/src/Main.purs
```diff
@@ -140,7 +139,11 @@ compile = do
         liftEffect cleanUpMarkers
 
         case res of
-          Right (CompileSuccess (SuccessResult { js, warnings })) -> do
+          Left err -> liftEffect do
```
thomashoneyman, line 142: This implementation has barely changed; I've just moved it from the end of the case statement to the top of the case statement.

client/src/Try/API.purs
```diff
@@ -19,111 +19,74 @@ import Affjax.RequestBody as AXRB
...
-import Foreign.Class (class Decode, decode)
-import Foreign.Generic (defaultOptions, genericDecode)
-import Foreign.Generic.Class (Options, SumEncoding(..))
-
-decodingOptions :: Options
-decodingOptions = defaultOptions { unwrapSingleConstructors = true }
 
 -- | The range of text associated with an error
-newtype ErrorPosition = ErrorPosition
+type ErrorPosition =
```
thomashoneyman, line 31: All of these types can be generically encoded and decoded via Argonaut without requiring newtypes and generic instances, so I've returned them to be raw records. However, if we really want those newtypes (for example, for more readable type errors) then I can reinstate them.

client/spago.dhall
hdgarrood: I’m not aware of any reason to keep dependencies here if we aren’t using them. 👍

server/Main.hs
```diff
@@ -2,6 +2,7 @@
 {-# LANGUAGE DeriveGeneric #-}
 {-# LANGUAGE LambdaCase #-}
 {-# LANGUAGE OverloadedStrings #-}
+{-# LANGUAGE PatternGuards #-}
```
thomashoneyman, line 5: I'm not sure why this has started happening, but CI fails on the server with:

```
[1 of 1] Compiling Main

/home/travis/build/purescript/trypurescript/server/Main.hs:136:27: error: [-Werror]

    accepting non-standard pattern guards (use PatternGuards to suppress this message)

        Just js <- codegenResult

    |

136 |                   Right _ | Just js <- codegenResult -> do

    |      
```

client/package.json
```diff
@@ -6,9 +6,9 @@
   },
   "scripts": {
     "clean": "rimraf output",
-    "test": "spago test --path $npm_package_config_configpath",
-    "build": "spago build --path $npm_package_config_configpath",
-    "bundle": "spago bundle-app --path $npm_package_config_configpath --purs-args '--censor-lib --strict' --to public/js/index.js"
+    "test": "spago test --path config/dev/Try.Config.purs",
+    "dev": "spago build --path config/dev/Try.Config.purs",
```
hdgarrood, line 10: If we are going to hardcode this path here, I think we should just hardcode the production config path too. How about having the dev one called `build` and the production one called `build_ci` or `build_production`?

client/src/Main.purs
```diff
@@ -43,7 +42,7 @@ displayErrors errs = do
   column2 <- JQuery.select "#column2"
   JQueryExtras.empty column2
 
-  forWithIndex_ errs \i (CompilerError{ message }) -> do
+  forWithIndex_ errs \i ({ message }) -> do
```
hdgarrood, line 45: These parens are unnecessary

client/src/Main.purs
```diff
@@ -149,43 +152,39 @@ compile = do
               sources <- runExceptT $ runLoader loader (JS js)
               liftEffect hideLoadingMessage
               for_ warnings \warnings_ -> liftEffect do
-                let toAnnotation (CompileWarning{ errorCode, position, message }) =
-                      position <#> \(ErrorPosition pos) ->
+                let toAnnotation ({ errorCode, position, message }) =
```
hdgarrood, line 155: Unnecessary parens here too

server/Main.hs
hdgarrood: Maybe we updated to a newer GHC version which produces this warning where it didn't previously?

hdgarrood (approved):

hdgarrood:
I think we can also remove the `config` key in package.json now too. Also it might be worth checking whether the changes to these package.json scripts mean that instructions in the readme or something need to be updated, if you haven’t done that already?

thomashoneyman:
Good point on the `config` key. The README doesn't need any changes as far as I can tell.

client/src/Try/API.purs
```diff
@@ -135,13 +98,13 @@ get url = ExceptT $ AX.get AXRF.string url >>= case _ of
...
   Left e ->
     pure $ Left $ printError e
   Right { status } | status >= StatusCode 400 ->
     pure $ Left $ "Received error status code: " <> show status
   Right { body } ->
-    pure $ Right $ runExcept (decode (unsafeToForeign body))
+    pure $ Right $ decodeJson body
   where
-  requestBody = AXRB.String code
+  requestBody = Just $ AXRB.Json $ encodeJson code
```
milesfrain, line 110: Wondering if the CORS issue reported in #215 is (unexpectedly) related to using `AXRB.Json` instead of `AXRB.String`.

thomashoneyman: Wow, great find -- it totally is. Changing this to `AXRB.String $ unsafeCoerec (encodeJson code)` works as expected. I don't know why -- going to look a little deeper.

thomashoneyman: Well, on reflection the code isn't actually JSON. It's just a string. Encoding it runs through a `stringify` call, which isn't what we want; I still don't know why this would cause a CORS issue (perhaps it's a red herring?) but regardless this wasn't a correct change to make in the first place.

PR #213, opened 2021-03-27
Update package set for PureScript 0.14

thomashoneyman:
Supersedes #205 by updating to a 0.14 package set. This package set should be a decrease in memory load because it's smaller than the later 0.13.x package sets (like the one currently running on Try PureScript). It's also necessary to update the package set to 0.14 before deploying a new version of Try PureScript using the 0.14 compiler.


server/Main.hs
```diff
@@ -2,6 +2,7 @@
 {-# LANGUAGE DeriveGeneric #-}
 {-# LANGUAGE LambdaCase #-}
 {-# LANGUAGE OverloadedStrings #-}
+{-# LANGUAGE PatternGuards #-}
```
thomashoneyman, line 5: CI fails otherwise.

JordanMartinez:
Just a thought. Could we not install the node libraries since they won't be
used anyway?

thomashoneyman:
We could blocklist libraries we don’t want to include, certainly. But if we can afford to keep libraries I think we should, even if it’s just so folks can share example code (even if it doesn’t run, it at least compiles).

hdgarrood (approved):

PR #215, opened 2021-03-28
Migrate to Halogen

thomashoneyman:
This PR continues #199 by replacing the existing jQuery with Halogen, except for the iframe -- the iframe deserves its own conversation and so I've left it untouched for this pull request. I have avoided changing any features and the application should look and behave the exact same as it currently does.

Well...there's _one_ difference: the old code tried to apply the Ace 'Dawn' theme, but it wasn't working; in this version the dawn theme is present. I can remove the theme altogether if desired, which I believe will default to the TextMate theme.

To review the code I would recommend pulling up the old `index.html` and old `Main.purs` files against the new `Container.purs` Halogen component, as the component wraps in functionality that was previously spread across those two files. To review the functionality I'd just make sure the editor works fine for you. I tested it by putting up the old editor and the new one side-by-side in browser tabs.

I've avoided doing some things that could be changed in the future if we want to clean things up a little:
  * I didn't introduce any routing; I instead still rely on the existing query string parsing
  * I didn't update to PureScript 0.14 / Halogen 6 (though I'd like to in the near future)
  * I only updated one line of CSS (the editor font size), but otherwise left it untouched.
  * I matched the Halogen HTML as exactly as possible to the old contents of index.html
  * I didn't introduce any new features or update any existing ones

My hope is that this unblocks some of the useful improvements @milesfrain provided a proof-of-concept for (like storing contents in the URL, re-enabling saving gists, and so on); with everything converted to Halogen those could be introduced one-by-one.

---

Here's a brief demonstration of the before-and-after. First, the current `try.purescript.org`:

https://user-images.githubusercontent.com/10245104/113072976-125daa80-917d-11eb-992c-2ef54f7f2fc8.mov

Next, the Halogen version on localhost:

https://user-images.githubusercontent.com/10245104/113072990-1a1d4f00-917d-11eb-968a-e07a508eac26.mov


milesfrain:
It looks like the CORS issue is introduced in #212.

For testing, I'm just pointing to:
- `https://compile.purescript.org` for compilation
- `localhost`/local path for serving all the prebuild .js files (which are built with `purs 0.13.8` for all tests, even when the client is sometimes built with 0.14).
  - That's accomplished with these [symlink setup steps](https://github.com/purescript/trypurescript/issues/183#issuecomment-650420721).

Here's a snapshot of a [branch](https://github.com/milesfrain/tps/tree/cors-bisect-master) with the last working commit before CORS issues. You can cherry-pick the next commit to trigger the failure. Best guess on a possible cause ([comment](https://github.com/purescript/trypurescript/pull/212/files#r603745443)).

thomashoneyman:
I followed up [on that comment](https://github.com/purescript/trypurescript/pull/212/files#r603745443) and that's absolutely the issue. I don't know why using a `Json` request body type causes a problem and using a `String` request body does not; I'll take a deeper look tomorrow, but if anyone else has an idea about this I'd appreciate ideas.

thomashoneyman:
I've updated the request in question and things appear to be working correctly now. Now that I'm able to run things properly I can test this out and make sure it's ready for review.

client/public/css/index.css
```diff
@@ -181,6 +181,7 @@ body {
   left: 0;
   right: 0;
   bottom: 0;
+  font-size: 13px;
```
thomashoneyman, line 184: Setting the font size on the editor container works fine; setting it via the Ace bindings is not currently working and I'll need to file a bug with `purescript-contrib/purescript-ace`.

client/public/index.html
```diff
@@ -1,197 +1,78 @@
...
+      $.ajaxSetup({
+        dataType: "text",
+      });
 
-    function cleanUpMarkers() {
-      for (var i = 0; i < cleanupActions.length; i++) {
-        cleanupActions[i]();
+      function teardownIFrame() {
+        var $ctr = $("iframe#output-iframe");
+        $ctr.remove()
```
thomashoneyman, line 29: This is necessary because several jQuery functions in PureScript were previously clearing the `column2` container (in order to clear the iframe) before displaying any contents in the column. Since the iframe is still here in jQuery I need a way to imperatively clear the iframe from Halogen.

client/src/Try/API.purs
```diff
@@ -98,13 +98,13 @@ get url = ExceptT $ AX.get AXRF.string url >>= case _ of
...
+compile endpoint code = ExceptT $ liftAff $ AX.post AXRF.json (endpoint <> "/compile") requestBody >>= case _ of
   Left e ->
     pure $ Left $ printError e
   Right { status } | status >= StatusCode 400 ->
     pure $ Left $ "Received error status code: " <> show status
   Right { body } ->
     pure $ Right $ decodeJson body
   where
-  requestBody = Just $ AXRB.Json $ encodeJson code
+  requestBody = Just $ AXRB.string code
```
thomashoneyman, line 110: This change is necessary to avoid the CORS issue that @milesfrain helped me track down.

client/src/Try/Container.purs
```diff
@@ -0,0 +1,469 @@
...
+import Try.Editor as Editor
+import Try.Gist (getGistById, tryLoadFileFromGist)
+import Try.Loader (Loader, makeLoader, runLoader)
+import Try.QueryString (getQueryStringMaybe)
+import Try.Session (createSessionIdIfNecessary, storeSession, tryRetrieveSession)
+import Try.Types (JS(..))
+import Web.HTML (window)
+import Web.HTML.Window (alert)
+
+type Slots = ( editor :: Editor.Slot Unit )
```
thomashoneyman, line 37: I've chosen to implement this as a container component for the page and a dedicated editor component for just the code editor. It makes it a little easier to focus just on the editor when implementing new features for it, or just focus on the container when dealing with the full page. I'm open to different architectures, however.

thomashoneyman:
If we agree to move this direction and this PR looks good, then I'd like to explore deploying a new version of Try PureScript for 0.14 before we start adding any new features.

hdgarrood:
> I followed up on that comment and that's absolutely the issue. I don't know why using a Json request body type causes a problem and using a String request body does not; I'll take a deeper look tomorrow, but if anyone else has ideas about this I'd appreciate hearing them.

I suspect that you were sending a JSON-encoded string to the server, e.g. the request might have looked like

```
POST /compile HTTP/1.1

"module Main where\nfoo = bar\n"
```

as opposed to what the server was expecting, which would have been

```
POST /compile HTTP/1.1

module Main where
foo = bar
```

This is where the "network" tab in the browser dev tools comes in handy.

client/src/Try/Editor.purs
```diff
@@ -0,0 +1,234 @@
...
+        let cleanup (MarkerId n) = EditSession.removeMarker n session
+        traverse_ cleanup markers
+
+    HandleChange -> do
+      let
+        handleChange = H.gets _.editor >>= traverse_ \editor -> do
+          text <- H.liftEffect (Editor.getValue editor)
+          H.raise $ TextChanged text
+
+      H.gets _.debounceRef >>= traverse_ \ref -> do
```
hdgarrood, line 142: Can we extract this debouncing logic out? It's a little complex, and I think currently more mixed up with the editor component logic than it needs to be. I also think it might be nice to consider implementing it in terms of `setTimeout` rather than `AVar`, like the lodash version is.

I'm imagining something like a function `debounce :: forall a. (a -> Aff Unit) -> Milliseconds -> Aff (a -> Aff Unit)` which would be run once when the component mounts; the outer `Aff` in the return value would be to allow the debouncing function to set up the state it needs.

hdgarrood (approved):
I'll leave it up to you to decide what you want to do about the debouncing. This looks good - I'm not a Halogen expert but the code is still relatively easy to follow, so nice work :+1:

client/src/Try/Editor.purs
thomashoneyman: It turns out this is just done in `Effect` anyways, so I could write a simple debouncer using `setTimeout` and `clearTimeout`. The other debouncer allows full evaluation in `HalogenM`, but you're right that it's overkill here. I've tested out the new implementation and it works fine.

PR #218, opened 2021-05-25
Support loading files from GitHub repositories

### LINKED ISSUE #164: Move example gists into the repo

thomashoneyman:
Supersedes #194 and #181 and fixes #164. This PR:

1. Implements a new query parameter, `github`, which accepts a path to a source file in a GitHub repository and loads its contents.
1. Moves all examples from their various gists into this repository under the `examples` directory
1. Fixes the generics example to be compatible with PureScript 0.14
1. Updates the default example to link to the GitHub files in this repository instead of the gists.

I also updated the 0.14 package set we use as a few new packages have been added since the last update.

This is the last change necessary before we can release Try PureScript for PureScript 0.14, in my opinion.


thomashoneyman:
I've tested and confirmed that this works as expected using the production build. I haven't been able to get the local dev server working properly, however -- I'm getting "Module Not Found" compilation errors for things that ought to be found, like `import Effect`. I'll confirm here that everything is fine with the dev server too once I can get that working.

README.md
```diff
@@ -12,9 +12,36 @@
...
-- Load PureScript code from Github Gists
+- Load PureScript code from GitHub Gists or repository files
 
-### Which Libraries are Available?
+### Control Features via the Query String
+
+Most of these features can be controlled not only from the toolbar, but also using the [query parameters](https://en.wikipedia.org/wiki/Query_string):
+
+- **Load From GitHub Repo**: Load PureScript code from a GitHub repository using the `github` parameter
+    - Example: `github=/purescript/trypurescript/master/client/examples/Main.purs` will load the code from this file. Note: the file should be a single PureScript module with the module name `Main`.
```
thomashoneyman, line 22: Only this section is changed; the rest of this diff is because I swapped this section to be higher (as it's more relevant to users than the "How to update the package set" section)

client/examples/Main.purs
```diff
@@ -0,0 +1,66 @@
...
+      , indent (p (code (text "  try.purescript.org?gist=gist-id")))
+      , p (fold
+          [ text "The Gist should contain PureScript modulenamed "
+          , code (text "Main")
+          , text "in a file named "
+          , code (text "Main.purs")
+          , text " containing your PureScript code."
+          ])
+      , p (text "To share code from a repository, include the path to the source file the URL as follows:")
+      , indent (p (code (text "  try.purescript.org?github=/owner/repo/Source.purs")))
```
hdgarrood, line 27: Do we need to include the branch name here?

client/examples/Main.purs
```diff
@@ -0,0 +1,66 @@
...
+    render $ fold
+      [ h1 (text "Try PureScript!")
+      , p (text "Try out the examples below, or create your own!")
+      , h2 (text "Examples")
+      , list (map fromExample examples)
+      , h2 (text "Share Your Code")
+      , p (text "A PureScript file can be loaded from GitHub from a gist or a repository. To share code using a gist, simply include the gist ID in the URL as follows:")
+      , indent (p (code (text "  try.purescript.org?gist=gist-id")))
+      , p (fold
+          [ text "The Gist should contain PureScript modulenamed "
```
hdgarrood, line 20:
```suggestion
          [ text "The Gist should contain a PureScript module named "
```

client/examples/Main.purs
```diff
@@ -0,0 +1,66 @@
...
+      , indent (p (code (text "  try.purescript.org?github=/owner/repo/Source.purs")))
+      , p (fold
+          [ text "The file should be a PureScript module named "
+          , code (text "Main")
+          , text " containing your PureScript code."
+          ])
+      ]
+  where
+    fromExample { title, source } =
+      link ("https://github.com/purescript/trypurescript/load-from-github/client/examples/" <> source) (text title)
```
hdgarrood, line 36: I think this ought to use a tag from some kind of configuration rather than hardcoding the branch name, so that we can change these without breaking the currently released version. In practice we will want this to be the tag from the current version.

thomashoneyman: Yea — I could thread this in as an argument to main, so that it can be accessed in various places taken from outside the PureScript code.

hdgarrood: Sounds good :+1:

client/examples/Main.purs
thomashoneyman: Yea, I suppose it ought to be /blob/<tag>/File.purs

client/examples/Main.purs
thomashoneyman: I could do that with the entire dev / prod configurations in that case

hdgarrood: If that makes the overall setup simpler, sure, I don't see why not.

thomashoneyman: On second thought, doing this would require some kind of text replacement processing pass to work. After all, this example is code loaded literally into the Try PureScript editor.

Otherwise, we'd have to do something like read the current tag from data in the index.html file or something like that, but this would be user-visible code and I'm not a huge fan of that.

Another option is to manually hard-code the tag as part of the release process and document where the tag would need to be updated -- we release so infrequently that this feels legitimately like an option.

hdgarrood: True - I think manually hardcoding the tag and documenting that it needs updating will work.

hdgarrood (approved):

client/examples/Main.purs
thomashoneyman: Perhaps we should split the repository README so that the main README is for users and there's a separate development guide. It could include instructions for developing, like we already have, but could also have a release guide which includes:

- Updating the hardcoded tag
- Checking externs sizes
- Running the deployment

What do you think?

thomashoneyman: For now I've just added a note about the hardcoded tags to the README, but I think we probably do want a dedicated, separate development + release guide.

PR #221, opened 2021-05-26
Update local compile server setup instructions so that code compiles

### LINKED ISSUE #220: Get local development server to work again
Currently, following the instructions [here](https://github.com/purescript/trypurescript#2-local-compile-server-setup) do not work. Further exploration has revealed that things break [in this line](https://github.com/purescript/trypurescript/blob/master/server/Main.hs#L233), which matches the same code used in the [REPL command](https://github.com/purescript/purescript/blob/master/app/Command/REPL.hs#L314-L323).

While this is related to #183 and #219 in that we need to update the instructions, this issue is specifically about figuring out why the server can't start.

JordanMartinez:
Fixes #220

### DIFF
--- a/README.md
+++ b/README.md
@@ -100,6 +100,11 @@ stack build
 cd staging
 spago install
 
+# Disable glob expansion to prevent 'ModuleNotFound' errors.
+# `purs` will handle that for us. See #220 for more contxt.
+#    Note: to re-enable, run `set +o noglob`
+set -o noglob
+
 stack exec trypurescript 8081 $(spago sources)
 # should output that is is compiling the sources (first time)
 # then: Setting phasers to stun... (port 8081) (ctrl-c to quit)
--- a/README.md
+++ b/README.md
@@ -100,12 +100,13 @@ stack build
 cd staging
 spago install
 
-# Disable glob expansion to prevent 'ModuleNotFound' errors.
-# `purs` will handle that for us. See #220 for more contxt.
-#    Note: to re-enable, run `set +o noglob`
-set -o noglob
-
-stack exec trypurescript 8081 $(spago sources)
+# Below, we disable glob expansion via `set -o noglob`
+# to prevent 'ModuleNotFound' errors. `purs` will handle
+# glob expansion correctly for us. See #220 for more contxt.
+#
+# We run this in a subshell so the change doesn't persist after
+# running the below command
+(set -o noglob && stack exec trypurescript 8081 $(spago sources))
 # should output that is is compiling the sources (first time)
 # then: Setting phasers to stun... (port 8081) (ctrl-c to quit)
 ```


README.md
```diff
@@ -100,6 +100,11 @@ stack build
 cd staging
 spago install
 
+# Disable glob expansion to prevent 'ModuleNotFound' errors.
+# `purs` will handle that for us. See #220 for more contxt.
+#    Note: to re-enable, run `set +o noglob`
+set -o noglob
```
hdgarrood, line 106: I think if you do `(set -o noglob && stack exec …)` that ought to still work and won’t affect the current shell, so you won’t need to re-enable globbing later.

f-f: A common alternative is also to `set +o noglob` after the `stack exec` command

hdgarrood: Right - what I’m saying is that I think it’s preferable not to have to do that. It’s annoying to have to turn it on and off if you’re running normal commands which do rely on globbing in between running the Try PureScript server

f-f: Sorry nvm, I thought this was a script, but it's dev instructions

JordanMartinez: Done.

thomashoneyman:
Hm -- even running `(set -o noglob && stack exec trypurescript 8081 $(spago sources))` still produces this when running a fresh clone of the repository and the dev server.

Note that the development instructions say to use `npm run build`, but this should probably say `npm run build:dev`.

Screen Shot 2021-05-26 at 9 43 55 PM

To clarify, I don't see the issue when compiling the `trypurescript` executable; it's only when making the POST request to the running dev server. Loading up `localhost:8080` is successfully making one POST request, but the compilation results in:

```json
{
  "error": {
    "tag": "CompilerErrors",
    "contents": [
      {
        "suggestion": null,
        "moduleName": "Main",
        "errorLink": "https://github.com/purescript/documentation/blob/master/errors/UnknownName.md",
        "errorCode": "UnknownName",
        "message": "  Unknown module Data.Foldable\n",
        ...
      }
    ]
  }
}
```

I see the same result whether I'm using `zsh` (the default for MacOS these days) or `bash`, so I don't think that's the issue. If these dev instructions work fine for everyone else  then perhaps this is just me and we should merge.

JordanMartinez:
I feel like that's a separate issue we can investigate after this PR gets merged. Beforehand, we had 'ModuleNotFound' errors, so that the server wouldn't even start. Now the server starts, but isn't working correctly. Perhaps the issue is now caused by a bug in the `server` function or that the project's dependencies just need to be updated to match `v0.14.1`'s dependency version?

thomashoneyman:
Ah, I see.

I was referring to the server starting, but compilation of the code provided via the client would fail with Module Not Found errors like the one I showed above. In my shell, at least, the dev backend comes up regardless of the noglob option being set.

But I’ve seen these errors stop the server from being compiled in the first place, so we could still merge this.

It shouldn’t be a 0.14.1 issue, because the package set in the staging directory has been updated to 0.14.1.

thomashoneyman (approved):

PR #223, opened 2021-05-28
Update trypurescript deps to match purescript repo versions

JordanMartinez:


trypurescript.cabal
```diff
@@ -13,23 +13,23 @@ author: Phil Freeman
...
 executable trypurescript
-    build-depends: base ==4.*,
-                   aeson -any,
-                   bytestring >=0.10.0.2 && <0.11,
-                   data-default -any,
-                   directory -any,
-                   filepath -any,
-                   Glob -any,
-                   scotty -any,
+    build-depends: base ==4.14.1.0,
```
hdgarrood, line 16: Can we leave the bounds as `-any` here, please? All of these dependencies are already sufficiently nailed down by `purescript` itself. The only one where I think we should keep the bounds is `scotty`.

stack.yaml
```diff
@@ -1,22 +1,18 @@
...
     assoc: false
-    quickcheck: false
 packages:
 - '.'
 extra-deps:
-# purescript 0.14.0-rc5
+# purescript 0.14.1
 - github: purescript/purescript
-  commit: 7ecc42669c69682996f2196ba2eef6c4ca827348
+  commit: 0d570f6e3828fc1a93c7c20252bccb04b7d2b0f6
```
hdgarrood, line 12: Can we switch back to using a released version of `purescript` now please?

trypurescript.cabal
JordanMartinez: Meaning, if a dependency version was listed as `-any`, keep it as that? Since your comment here was made on the `base` line, I'm getting conflicting information in this comment.

stack.yaml
JordanMartinez: Yeah. One downside of this is that `purescript` depends on `purescript-ast`, which is going to be merged into `purescript-cst`. So, we'll need to update this again to drop `purescript-ast` once `v0.14.2` is released.

hdgarrood: That's fine, we'll need to update this file at that point anyway.

trypurescript.cabal
hdgarrood: Yeah sorry, I think `base` can be left alone too.

JordanMartinez: I've force-pushed some changes to reset earlier work. What needs to change now?

stack.yaml
```diff
@@ -1,22 +1,14 @@
-resolver: lts-13.26
+resolver: lts-17.6
 flags:
   aeson-pretty:
     lib-only: true
   these:
     assoc: false
-    quickcheck: false
```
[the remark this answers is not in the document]
JordanMartinez: It seems like it's an invalid flag. If I run `stack build` with that included, I get this error:
```
$ stack build
Invalid flag specification:
- Package 'these' does not define the following flags (specified in stack.yaml):
  quickcheck
- Flags defined by package 'these':
  these:assoc
```

Note: [PureScript's `stack.yaml`](https://github.com/purescript/purescript/blob/master/stack.yaml#L29-L31) file doesn't have that flag either.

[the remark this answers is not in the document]
hdgarrood: Oh right ok sorry, I didn’t realise it had been removed.

hdgarrood (approved):

trypurescript.cabal
```diff
@@ -15,21 +15,21 @@ data-dir: ""
...
                    data-default -any,
                    directory -any,
                    filepath -any,
                    Glob -any,
-                   scotty -any,
+                   scotty ==0.12,
                    purescript,
                    purescript-cst,
                    containers -any,
                    http-types >= 0.8.5,
```
JordanMartinez, line 27: @hdgarrood Before I merge this, should `http-types` be updated to `-any` or to an the [`lts-17.6` package set](https://www.stackage.org/lts-17.6)'s version, which is `0.12.3`?

And is it correct that `-any` will use whichever dependencies work with `purescript` and the other versions dependencies? I'm not familiar with this syntax unfortunately.

hdgarrood: Thinking about this a little more I think we might as well use `-any` for all of the dependencies here. This project is only ever going to be built by `stack`, and the stack.yaml file already pins everything down to a specific version, so additionally specifying versions here doesn’t really add anything.

It’s more important to provide accurate version bounds when you’re making a library which other people are depending on - this part of the cabal file does the same job as the dependencies part of a bower.json or package.json file in a JS library. But that’s not what we are doing here so it doesn’t really matter.

hdgarrood (approved):

PR #224, opened 2021-05-28
Provide externs to rebuildModule

### LINKED ISSUE #222: Unable to use modules from package set
While #220 made the server work, the larger goal was to make this entire project work in a local setting. So, this issue seeks to fix that problem.

The current issue, [explained by Thomas here](https://github.com/purescript/trypurescript/pull/221#issuecomment-849318543), is that modules aren't known when we attempt to compile code that uses them.

thomashoneyman:
Closes #222. When implementing #209 I accidentally provided an empty array to `rebuildModule` for the externs; this fix properly threads the externs through. As @JordanMartinez noticed in #222 we also can remove the `initNamesEnv` argument as it is no longer used.

Evidence that things work once again:

Screen Shot 2021-05-28 at 1 46 10 PM

### DIFF
--- a/server/Main.hs
+++ b/server/Main.hs
@@ -112,8 +112,8 @@ buildMakeActions codegenRef =
   outputPrimDocs :: Make.Make ()
   outputPrimDocs = pure ()
 
-server :: [P.ExternsFile] -> P.Env -> P.Environment -> Int -> IO ()
-server externs initNamesEnv initEnv port = do
+server :: [P.ExternsFile] -> P.Environment -> Int -> IO ()
+server externs initEnv port = do
   codegenRef <- IORef.newIORef Nothing
   let makeActions = buildMakeActions codegenRef
   let compile :: Text -> IO (Either Error ([P.JSONError], JS))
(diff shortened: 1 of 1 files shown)


JordanMartinez (approved):

hdgarrood (approved):
This is definitely a step in the right direction, but I’m curious about a) whether initEnv is used, and b) if we can get unused bindings to emit errors which cause CI to fail. Ideally this would have been caught before it was merged in the first place.

JordanMartinez:
> whether initEnv is used

It's used in the [`/search` path for doing searches](https://github.com/purescript/trypurescript/blob/master/server/Main.hs#L165-L166)

JordanMartinez:
> if we can get unused bindings to emit errors which cause CI to fail

This sounds like a separate PR to me. Perhaps we should follow the hlint warnings that `rhendric` added to the `purescript` repo recently?

hdgarrood:
No need to bring hlint in for that - GHC is already capable of generating warnings for unused bindings, and then -Werror / —pedantic will cause the build to fail if there are warnings.

hdgarrood:
By the way, we could also keep the initNamesEnv around and use `P.rebuildModule'` here, as that one allows you to provide a prebuilt Env rather than having to construct it each time, as that would speed the server up a little bit. See https://github.com/purescript/purescript/blob/master/src/Language/PureScript/Make.hs, specifically how `rebuildModule` is the same as the primed version except that it constructs an Env out of the externs first.

JordanMartinez:
> By the way, we could also keep the initNamesEnv around and use P.rebuildModule' here

I would but [`rebuildModule'` is not exported](https://github.com/purescript/purescript/blob/master/src/Language/PureScript/Make.hs#L1-L9). Not sure if that's intentional or not. That was my initial thought when I saw how similar that was to `rebuildModule`.

hdgarrood:
Ah, I see, thanks. I think that might be unintentional. It would definitely be useful in cases like this one when you’re using the compiler from Haskell.

JordanMartinez:
I've submitted a PR to export it.

PR #225, opened 2021-05-28
Update development instructions

thomashoneyman:
Closes #202, #196, #183. Confirmed that on a fresh clone of Try PureScript, following these instructions exactly, I can run the client against a local server or the production Try PureScript server with no additional tools or steps. cc: @milesfrain.

PR #226, opened 2021-05-29
Create valid symbolic link to output dir

JordanMartinez:
I was getting a 500 error in [#225 (comment)](https://github.com/purescript/trypurescript/pull/225#issuecomment-850724476) because the symbolic link created via `ln -s output ../client/public/js/output` was broken. 

Adding the `--relative` flag fixes the issue for me. I get the same view Thomas did in #225.

### DIFF
--- a/README.md
+++ b/README.md
@@ -102,9 +102,7 @@ spago build
 # Note: due to the differences in the `ln` command between
 # MacOS (BSD ln) and Linux (GNU ln), the below code will work
 # on both operating systems:
-pushd ../client/public/js
-ln --symbolic ../../../staging/output/ .
-popd
+ln -s "$PWD/output" "$PWD/../client/public/js/output"
 
 # Then, start the server.
 #
--- a/README.md
+++ b/README.md
@@ -99,9 +99,6 @@ cd staging
 spago build
 
 # Ensure the compiled JavaScript is available to the client via symbolic link.
-# Note: due to the differences in the `ln` command between
-# MacOS (BSD ln) and Linux (GNU ln), the below code will work
-# on both operating systems:
 ln -s "$PWD/output" "$PWD/../client/public/js/output"
 
 # Then, start the server.
--- a/README.md
+++ b/README.md
@@ -99,7 +99,7 @@ cd staging
 spago build
 
 # Ensure the compiled JavaScript is available to the client
-ln -sr output ../client/public/js/output
+ln --symbolic --relative output ../client/public/js/output
 
 # Then, start the server.
 #
--- a/README.md
+++ b/README.md
@@ -99,6 +99,13 @@ cd staging
 spago build
 
 # Ensure the compiled JavaScript is available to the client
+#   The `ln` command works differently between MacOS and Linux.
+#   Use the command below that corresponds to your operating system:
+
+#   MacOS users
+ln --symbolic output ../client/public/js/output
+
+#   Linux users (requires the `--relative` flag)
 ln --symbolic --relative output ../client/public/js/output
 
 # Then, start the server.
--- a/README.md
+++ b/README.md
@@ -98,15 +98,13 @@ stack build
 cd staging
 spago build
 
-# Ensure the compiled JavaScript is available to the client
-#   The `ln` command works differently between MacOS and Linux.
-#   Use the command below that corresponds to your operating system:
-
-#   MacOS users
-ln --symbolic output ../client/public/js/output
-
-#   Linux users (requires the `--relative` flag)
-ln --symbolic --relative output ../client/public/js/output
+# Ensure the compiled JavaScript is available to the client via symbolic link.
+# Note: due to the differences in the `ln` command between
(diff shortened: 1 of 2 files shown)


hdgarrood:
Huh, interesting. What platform are you on? CLI tools sometimes have different behaviour across platforms, especially if they have been around a long time, so it would probably be good to check that this works in both macOS and Linux.

JordanMartinez:
Here's my system info:
```
Linux Kernel: 5.8.0-44-generic
OS: Linux Mint 20.1
GNU bash, version 5.0.17(1)-release (x86_64-pc-linux-gnu)
ln (GNU coreutils) 8.30
```

hdgarrood:
Ok, thanks. Generally the tools on macOS systems are more similar to BSD versions, since that’s what mac is based on, and unfortunately it looks like the BSD version of ln doesn’t support long flags or the —relative flag: https://www.freebsd.org/cgi/man.cgi?ln

I think the most portable option might be to use $PWD so that the links don’t have to be relative, something like `ln -s output $PWD/../client/public/js/output` perhaps?

JordanMartinez:
I found this comment that says [`-r` is a GNU extension](https://unix.stackexchange.com/questions/153015/why-ln-s-creates-relative-broken-links#comment1116370_153016) and another comment saying [MacOS doesn't have the `-r` option](https://unix.stackexchange.com/questions/10370/make-a-symbolic-link-to-a-relative-pathname/10371#comment488941_152088). 

Let me try your idea.

JordanMartinez:
Nope. That doesn't work for me. Perhaps it would be easier to just note the two differences?

hdgarrood:
What’s the error? I’m sure there’s a way of getting it to work in both systems.

JordanMartinez:
```bash
# cwd = `trypurescript/staging`
$ ln -sv output $PWD/../client/public/js/output
'/home/jordan/Programming/Projects/trypurescript/staging/../client/public/js/output' -> 'output'

# cwd = `trypurescript/client/public/js`
$ file output/Data.Foldable/index.js
output/Data.Foldable/index.js: cannot open `output/Data.Foldable/index.js' (Too many levels of symbolic links)
```

hdgarrood:
Right ok, I think what you have there is the `client/public/js/output` directory is a symbolic link that points to itself. I think the first argument is being interpreted relative to the second one. Maybe try `ln -s “$PWD/output” “$PWD/../client/js/output”`?

JordanMartinez:
```bash
$ ln -s “$PWD/output” “$PWD/../client/js/output”
ln: failed to create symbolic link '“/home/jordan/Programming/Projects/trypurescript/staging/../client/js/output”': No such file or directory
```

JordanMartinez:
:laughing: This solution works, but only because it completely removes the relative path in the second argument
```bash
cd client/public/js
ln -s ../../../staging/output/ .
```

hdgarrood:
The “no such file or directory” error there is probably due to the fancy quotes, you’ll want to use normal ones

JordanMartinez:
You're right. I copy-pasted what you had above.

@thomashoneyman Can you confirm that this produces a valid symbolic link on your side?
```bash
ln -s "$PWD/output" "$PWD/../client/public/js/output"
```

thomashoneyman:
Yep! Works for me. Though I don't need the quotes -- `$PWD/output $PWD/../client/public/js/output` works just fine as well. Either way is good.

thomashoneyman (approved):

README.md
```diff
@@ -98,8 +98,11 @@ stack build
 cd staging
 spago build
 
-# Ensure the compiled JavaScript is available to the client
-ln -s output ../client/public/js/output
+# Ensure the compiled JavaScript is available to the client via symbolic link.
+# Note: due to the differences in the `ln` command between
+# MacOS (BSD ln) and Linux (GNU ln), the below code will work
+# on both operating systems:
```
thomashoneyman, line 104: We could probably omit this and just expect that folks copy/paste it.

JordanMartinez:
> Though I don't need the quotes -- $PWD/output $PWD/../client/public/js/output works just fine as well. Either way is good.

I need the quotes. Otherwise, it doesn't work.

README.md
JordanMartinez: Could, but are we going to remember why this was needed? Do we need to remember?

hdgarrood:
You'll need the quotes if you have spaces in any of the parent directories, such as if your username has a space in it.

README.md
hdgarrood: I do think this comment is a little bit odd - it's not due to the differences between the versions of `ln` that this command works on both mac and linux, it's _despite_ them! This command probably isn't going to change very often, and if it does break for one of the two OSes it's not the end of the world.

JordanMartinez: Ok. I'll remove since both of you think it's not needed.

thomashoneyman (approved):

PR #256, opened 2021-09-19
Adds longer loading screen

mikesol:
**Description of the change**

Keeps the loader up longer, lowering blank-screen time to less than 50ms.

---

**Checklist:**

### DIFF
--- a/client/src/Try/Container.purs
+++ b/client/src/Try/Container.purs
@@ -12,9 +12,10 @@ import Data.FoldableWithIndex (foldMapWithIndex)
 import Data.Maybe (Maybe(..), fromMaybe, isNothing)
 import Data.Symbol (SProxy(..))
 import Effect (Effect)
-import Effect.Aff (Aff)
+import Effect.Aff (Aff, makeAff)
+import Effect.Aff as Aff
 import Effect.Class.Console (error)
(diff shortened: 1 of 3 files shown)


JordanMartinez (approved):

client/src/Try/Container.purs
```diff
@@ -85,7 +86,9 @@ _editor = SProxy
 loader :: Loader
 loader = makeLoader Config.loaderUrl
 
-foreign import setupIFrame :: EffectFn1 (Object JS) Unit
+type LoadCb = Effect Unit
+type FailCb = Effect Unit
```
JordanMartinez, line 90: I don't think these type aliases are needed, but :shrug:

JordanMartinez:
@mikesol Could you add an entry to the changelog file?

client/src/Try/Container.purs
thomashoneyman: I'm OK with them in this case because it can be otherwise difficult to tell which is the 'load' callback and which is the 'fail' callback. In other projects I also use records to help, ie. `{ fail :: Effect Unit, load :: Effect Unit }`.

thomashoneyman (approved):

thomashoneyman (approved):

PR #259, opened 2021-11-13
Update package set to 0.14.5; doc package set size

JordanMartinez:
**Description of the change**

Updates package set to latest version.

Fixes changelog due to previous release.

Add instructions for calculating size of package set.

```
$ du --human-readable --max-depth=0 output/
185M	output/
```

---

**Checklist:**


thomashoneyman:
I'm not totally sure how we should be calculating the size of the package set; I've installed better metrics for Digital Ocean and I can see that we're at about 35% memory load (of 4GB RAM, so about 1.4GB of memory usage). We need to make sure that large package sets don't overwhelm the memory on the machine, but it's not clear to me how a 185MB package set corresponds to the 1.4GB in memory usage we have.

I'd like to better understand that relationship -- perhaps @hdgarrood knows a bit better how to quantify this? -- before specifying any kind of limit for the size of the package set.

That said, the rest of the instructions here are fine with me.

thomashoneyman:
If you'd like to just get the package set update merged and deployed, then you could do that in a new branch; since we're at 35% utilization, adding a handful of new packages won't be an issue. My review comments here are just about the changes in the instructions.

hdgarrood:
Yeah, I think it's probably not safe to make any assumptions about the amount of memory the server will need while running based on size of the `output` directory. I think the best thing to do is probably to run the server locally and then see how much memory it is using with a tool like `ps`. Or maybe I was being too cautious before, and you could just try deploying it and roll back if there are problems? I guess the worst case is a little bit of downtime, which is not really that bad.

JordanMartinez:
See #260

PR #270, opened 2022-02-06
[WIP] Add auto-complete suggestions

### LINKED ISSUE #32: Autocomplete
Ace probably provides an API for this, we just need to expose it on the server.

pete-murphy:
**Description of the change**

Fixes #32.

Clearly and concisely describe the purpose of the pull request. If this PR relates to an existing issue or change proposal, please link to it. Include any other background context that would help reviewers understand the motivation for this PR.

---

**Checklist:**

### DIFF
--- a/server/Main.hs
+++ b/server/Main.hs
@@ -20,7 +20,7 @@ import           Data.Aeson ((.=))
 import           Data.Bifunctor (first, second, bimap)
 import qualified Data.ByteString.Lazy as BL
 import           Data.Default (def)
-import           Data.Function (on)
+import           Data.Function (on, fix)
 import qualified Data.IORef as IORef
 import           Data.List (nubBy)
 import qualified Data.List.NonEmpty as NE
@@ -43,10 +43,12 @@ import qualified Language.PureScript.Make as Make
 import qualified Language.PureScript.Make.Cache as Cache
 import qualified Language.PureScript.TypeChecker.TypeSearch as TS
 import qualified Network.Wai.Handler.Warp as Warp
+import qualified System.Directory as Directory
 import           System.Environment (getArgs)
 import           System.Exit (exitFailure)
 import           System.FilePath.Glob (glob)
 import qualified System.IO as IO
+import qualified System.Process as Process
 import           Web.Scotty
 import qualified Web.Scotty as Scotty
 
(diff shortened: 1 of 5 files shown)


server/Main.hs
```diff
@@ -157,6 +181,35 @@ server externs initNamesEnv initEnv port = do
...
+                                  [ "search" .= q ]
+                                ]
+                              ]
+                            )
+                          ]
+                        ]
+      (Just handleIn, Just handleOut, _, _) <- liftIO ideClient
+      liftIO $ Char8.hPutStrLn handleIn (mkCommand (query :: Text))
+      result <- liftIO $ BS.hGetContents handleOut
+      Scotty.text (TL.fromStrict (T.decodeUtf8 result))
```
ptrfrncsmrph, line 212: This currently works, but there might be some inefficiency here that I'm not spotting. Running `purs ide client` directly (talking to same `purs ide server` instance) is instantaneous, while hitting this endpoint takes around 2s, sometimes longer.

ptrfrncsmrph: Maybe this could be a POST request with JSON constructed on the client, and passed directly to `purs ide client`

ptrfrncsmrph: It's somehow much faster using POST https://github.com/purescript/trypurescript/pull/270/commits/749670093a22c5b2abaee654a9e2051be302fee1, but I don't know if there's any security issues with just proxying directly to `purs ide server`

thomashoneyman:
@nwolverson -- I'm curious if you have any thoughts on this from the language server side of things; I'm not really sure what is and isn't possible with `psc-ide` to make this work.

nwolverson:
As per discussion on discord I'd suggest avoiding proxying the `purs ide` requests directly simply for the reason that it likely allows read/write access to the entire filesystem as the current user, and probably allows more significant DoS potential than a more directed endpoint. However returning an unparsed blob of json to be parsed on the client side seems like a legitimate option if it avoids re-encoding results & can reuse code.

It is surprising that the request would take 2s, it would seem the only extra overhead is an HTTP request, you are spawning the `purs` binary from the shell as well as the Haskell code. Are you actually looking at the same request in the same workspace? I see `groupReexports` and `maxResults` on the left, I'd also prefer to see a better timing (eg just directly using `time`).
Certainly making the socket request that `purs ide client` makes directly could be faster but that's something to shave off milliseconds I'd assume.

Happy to discuss the details of the completion results themselves later - but that request looks sane at least

ptrfrncsmrph:
This is surprising to me—these are the results from this commit https://github.com/purescript/trypurescript/pull/270/commits/0bdd1fd65505807b0f8fd3dabf5048aac5a1bec2
```
❯ hyperfine --warmup 1 "curl -Gs --data-urlencode 'q=app' localhost:8081/complete"
Benchmark #1: curl -Gs --data-urlencode 'q=app' localhost:8081/complete
  Time (mean ± σ):     173.3 ms ±  10.2 ms    [User: 7.3 ms, System: 11.3 ms]
  Range (min … max):   155.2 ms … 188.8 ms    16 runs
```
but removing this line https://github.com/purescript/trypurescript/blob/0bdd1fd65505807b0f8fd3dabf5048aac5a1bec2/server/Main.hs#L191 seems to explain the 2s slowdown
```
❯ hyperfine --warmup 1 "curl -Gs --data-urlencode 'q=app' localhost:8081/complete"
Benchmark #1: curl -Gs --data-urlencode 'q=app' localhost:8081/complete
  Time (mean ± σ):      2.339 s ±  0.118 s    [User: 5.0 ms, System: 8.3 ms]
  Range (min … max):    2.209 s …  2.590 s    10 runs
```
I added that line because the previous iteration with the POST request was hanging and that seemed to fix it, but I don't fully understand the behavior there.

Issue #264, opened 2022-01-18
Proposal: Introduce `esbuild` for Try PureScript


thomashoneyman:
The PureScript compiler is going [to switch to ES modules in 0.15](https://github.com/purescript/purescript/pull/3791), most likely [removing the `purs bundle` command](https://github.com/purescript/purescript/issues/4226). This will [affect the `spago bundle` command](https://github.com/purescript/spago/issues/836) as well, which will probably either be removed or will call out to a bundler on the user's behalf.

Try PureScript has historically been designed to work with only stock PureScript tooling and no external tools. However, doing that with the ES modules compiler means serving an unbundled application (unless `spago bundle` integrates a bundler for you, in which case we can use that). While an unbundled deployment is possible, [ES modules are significantly less performant than the same code bundled](https://v8.dev/features/modules#performance), and I'd rather just bundle.

In addition, we may wish to bundle SystemJS if we go forward with #263, and [newer versions of Ace have `require` statements for the Ace library](https://github.com/purescript-contrib/purescript-ace/blob/d185a5fa0b74f7edcd44b0ef16f37aea3868f567/src/Ace.js#L3) that also necessitate a bundler. We would either need to revert that change or introduce a bundler to Try PureScript to continue using Ace.

At the moment we have workarounds like making Ace available via a CDN so we can use it, but since we're likely to bundle in the future once the compiler emits ES modules, and we're likely to bundle if we adopt SystemJS, and we're going to have to update Ace to a version that requires bundling when we update to PureScript 0.15, I feel like we should just go ahead and introduce a bundler to Try PureScript.

My personal preference is `esbuild` (I can expand on why, if desired), and I would prefer to keep to either no configuration or an extremely minimal configuration -- the sort of thing that would replace `spago bundle`.

Issue #263, opened 2022-01-18
Proposal: Load foreign code with ES Module Shims and import maps


thomashoneyman:
Try PureScript is going to have to be reworked when the PureScript compiler [switches from CommonJS to ES modules](https://github.com/purescript/purescript/pull/3791) for foreign code. This issue details a proposal for how to handle this rewrite.

### Current Implementation

We currently parse through the generated JavaScript on Try PureScript to discover all dependencies present via `require()` statements. For every dependency we find we do one of:

1. Load the JS from `compile.purescript.org` (for example, [`Data.EuclidianRing/foreign.js`](https://compile.purescript.org/output/Data.EuclideanRing/foreign.js))
2. Look up its URL from the [Shim.purs](https://github.com/purescript/trypurescript/blob/master/client/src/Try/Shim.purs) file which lists CDN locations for some common FFI libraries like `react`, and then load the contents

Once we know what dependencies to load, we take their contents and insert them into the code iframe. Finally, we load their contents via the [`evalSources` function in `frame.js`](https://github.com/purescript/trypurescript/blob/af70d48e3d500ab6909e782b7ac50f982b6948e2/client/public/js/frame.js#L2-L28).

### Issues

This code is no longer going to work. PureScript 0.15 will only work with ES modules, so the code in `Shim.purs`, `Loader.purs`, and `frame.js` will all need to be updated to work with ES modules instead of CommonJS modules -- in other words, there is no more `require` or `module.exports`; instead, there's `import` and `export`.

I also don't think we can instantiate the code the same way we do now, which I _believe_ happens here:

https://github.com/purescript/trypurescript/blob/53a82a636802aad4007a0a6cbee9bc7dfdd1be4b/client/public/js/frame.js#L25

Finally, if we swap all the way over to ES modules, we lose all Internet Explorer support, which we have historically not wanted to lose.

### Proposal: SystemJS

We're clearly going to have to rewrite how we handle foreign code in Try PureScript. After a brief discussion with @natefaubion on the PureScript chat I would like to propose that we use two technologies:

1. [SystemJS](https://github.com/systemjs/systemjs), which [covers IE11](https://github.com/systemjs/systemjs#ie11-support)
2. [Import maps](https://wicg.github.io/import-maps), which are [supported in SystemJS](https://github.com/systemjs/systemjs#import-maps), including [for IE11](https://github.com/systemjs/systemjs#note-on-import-maps-support-in-ie11)

[...]

PR #274, opened 2022-04-30
Update to 0.15.0 PureScript

### LINKED ISSUE #273: Update to 0.15.0 compiler
Now that `0.15.0` is released, we can update this project. Not sure if we want to wait a bit until the package set gets filled with more packages first, but I imagine that will happen quickly over the next week.

JordanMartinez:
**Description of the change**

Fixes #273. Still a WIP. Besides updating the server code, I also updated the client and staging code to `0.15.0`.

---

**Checklist:**


client/packages.dhall
```diff
@@ -1,22 +1,14 @@
...
-          , "nullable"
-          , "prelude"
-          , "refs"
-          , "web-html"
-          , "web-uievents"
-          ]
+          [ "console", "effect", "functions", "prelude" ]
         }
-    }
+      }
```
JordanMartinez, line 12: This should be removed once `debug` gets added to the package set.

client/spago.dhall
```diff
@@ -38,8 +46,9 @@
   , "transformers"
   , "tuples"
   , "unfoldable"
+  , "unsafe-coerce"
   , "web-html"
   ]
 , packages = ./packages.dhall
-, sources = [ "src/**/*.purs", "test/**/*.purs" ]
+, sources = [ "src/**/*.purs", "config/dev/**/*.purs", "test/**/*.purs" ]
```
JordanMartinez, line 53: Perhaps we should use 3 `spago.dhall` files: a shared one, one for dev and one for prod?

mikesol:
This looks really good! I tried to hack at it a bit and spotted that `setupIFrame` and `teardownIFrame` are exported from Container.js without being defined there, which may be problematic for `esbuild`. Is that the bundler you're going to use for the `client`?

mikesol:
@JordanMartinez if you want to assign me as a reviewer I can mark the places where I needed to make changes for my dockerized version of trypurescript (the one on try.wags.fm). Some of them will be needed for this PR as well (for example, I hit the `esbuild` issues with `top` being a reserved word, etc).

JordanMartinez:
@mikesol This PR will need to update the code to do both #263 and #264.

> `setupIFrame` and `teardownIFrame` are exported from Container.js without being defined there, which may be problematic for esbuild

Looks like those two functions are defined in `client/public/index.html`. Since we'll be using SystemJS, I don't think we'll implement the code in the same way as before, so this might be a moot point.

JordanMartinez:
Succeeded by #275

PR #286, opened 2022-07-13
Use div-root; fix examples' warnings

### LINKED ISSUE #283: How to deal with warnings in examples?
This is a minor issue. I was wondering if we can or should address warnings shiowing up in the [default examples](https://github.com/purescript/trypurescript/tree/master/client/examples).

## Warnings

image

image

image

## Options

* Keep the examples as they are and show the warnings
* Change the code examples (e.g. add type annotation to `main`)
* Change Try-PureScript to hide these specific warnings (main type signature, TryPureScript import)

JordanMartinez:
**Description of the change**

Fixes #283. Also allows updates the `frame.html` file to include a `<div id="root"></div>`. See https://github.com/JordanMartinez/purescript-cookbook/issues/198

---

**Checklist:**

### DIFF
--- a/client/public/frame.html
+++ b/client/public/frame.html
@@ -47,5 +47,6 @@
 </head>
 <body>
   <main id="main"></main>
+  <div id="root"></div>
 </body>
 </html>
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,8 @@ New features:
 Bugfixes:
 
 Other improvements:
+- Fixes compiler warnings in examples (#286 by @JordanMartinez)
+- Support cookbook repo UI recipes by adding element to `frame.html` (#286 by @JordanMartinez)
 
 ## [v2022-07-12.1](https://github.com/purescript/trypurescript/releases/tag/v2022-07-12.1)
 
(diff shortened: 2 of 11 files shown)


client/public/frame.html
```diff
@@ -47,5 +47,6 @@
 </head>
 <body>
   <main id="main"></main>
+  <div id="root"></div>
```
thomashoneyman, line 50: I'm not totally following why this is required. Is the cookbook unable to use `main id="main"`? If not, and it seems important that we specifically use `<div id="root">`, then can we adjust the existing code (ie. the TryPureScript rendering) to point at this div and remove `<main id="main">`? It's a bit odd to have both.

JordanMartinez: `main` is used to redirect `Console.log` (when under a `TryPureScript.render =<< TryPureScript.withConsole`, so that the strings logged to the console are instead added as nodes under the `main id="main"` HTML element rather than logging to the web browser's console.

I don't want the two to mix, so I'm creating a separate element for React usage.

thomashoneyman (approved):

PR #290, opened 2022-07-16
Fix NPM shim resolver on Firefox

### LINKED ISSUE #289: NPM dependency shims no longer work on Firefox 102.1?
I think Firefox made a few changes recently (not exactly sure what) that breaks `es-module-shims`. Running the below code on https://try.purescript.org doesn't work for Firefox 102.1 but does on Chromium 103.0.5060.53. Can anyone else confirm?

```purs
module Main where

import Prelude

import Data.Foldable (fold)
import Effect (Effect)
import Effect.Console (log)
import TryPureScript as TP
import Data.Decimal as D

main :: Effect Unit
main = TP.render =<< TP.withConsole do
  log $ show $ D.fromString "132.4242"
```

JordanMartinez:
**Description of the change**

Fixes #289. However, trying to run the React examples from the cookbook get a weird error even with this fix:
- Works: [`HelloReactHooks`](https://github.com/JordanMartinez/purescript-cookbook/blob/master/recipes/HelloReactHooks/src/Main.purs)
- Fails: [`ButtonsReactHooks`](https://github.com/JordanMartinez/purescript-cookbook/blob/master/recipes/ButtonsReactHooks/src/Main.purs)

---

**Checklist:**

### DIFF
--- a/client/public/frame.html
+++ b/client/public/frame.html
@@ -41,7 +41,7 @@
   </script>
 
   <!-- ES Module Shims: Import maps polyfill for modules browsers without import maps support (all except Chrome 89+) -->
-  <script async src="https://ga.jspm.io/npm:es-module-shims@1.5.6/dist/es-module-shims.js" integrity="sha384-Zi3OSEiS8YnN0yIdnkInyBlzXCGA18AMnQuPb+T1hXvE4DIDhGIp2QIBZcm4WNIU" crossorigin="anonymous"></script>
+  <script async src="https://ga.jspm.io/npm:es-module-shims@1.5.9/dist/es-module-shims.js" integrity="sha384-O6YhhDNmwzHXz9VDaEaEtY9rnzCF5Fy+yoxD6StrNyhjwyBrYYOjZkbFRt9zcoGx" crossorigin="anonymous"></script>
 
   <script src="js/frame.js"></script>
 </head>
--- a/client/public/frame.html
+++ b/client/public/frame.html
@@ -18,23 +18,24 @@
   </script>
   <!--
     JSPM Generator Import Map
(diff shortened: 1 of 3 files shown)


JordanMartinez:
Looks like the import map for `react-dom/server` (https://ga.jspm.io/npm:react-dom@16.13.1/server.browser.js) returns this JS:
```js
import"react";import"object-assign";export{default}from"./cjs/react-dom-server.browser.production.min.js";
```
I'll investigate.

JordanMartinez:
The import map generator allows you to use either a "production" or "development" version of the dependencies. For our import maps, the mode only affects the react libraries. Previously, we were on the 'production' version. The last commit switches to the 'development' version. Since the development version uses non-minified JS code, identifiers like `createRoot` still exist rather than being minified to something like `e`.

With the latest commit, the `ButtonReactHooks` example now works:

client/public/frame.html
```diff
@@ -18,23 +18,24 @@
...
   {
     "imports": {
       "big-integer": "https://ga.jspm.io/npm:big-integer@1.6.51/BigInteger.js",
       "decimal.js": "https://ga.jspm.io/npm:decimal.js@10.3.1/decimal.js",
-      "react": "https://ga.jspm.io/npm:react@16.13.1/index.js",
-      "react-dom": "https://ga.jspm.io/npm:react-dom@16.13.1/index.js",
-      "react-dom/server": "https://ga.jspm.io/npm:react-dom@16.13.1/server.browser.js",
+      "react": "https://ga.jspm.io/npm:react@16.13.1/dev.index.js",
+      "react-dom": "https://ga.jspm.io/npm:react-dom@16.13.1/dev.index.js",
+      "react-dom/server": "https://ga.jspm.io/npm:react-dom@16.13.1/dev.server.browser.js",
```
ptrfrncsmrph, line 30: I wonder, does bumping the version also fix this? I would think this might need to be in sync with the `purescript-react` libraries which were recently updated to 18 (I'm actually surprised that `createRoot` is in the `dev` version of `16.13.1`).

JordanMartinez: Looks like you're right. Bumping the React version to `17.0.2`, the last release before 18, makes it work.

JordanMartinez:
@thomashoneyman Can I get a review on this when you get a chance?

thomashoneyman (approved):

PR #295, opened 2022-08-12
Fix double main invocation using shim mode

JordanMartinez:
**Description of the change**

Fixes #293.

---

**Checklist:**

- [x] Added the change to the changelog's "Unreleased" section with a reference to this PR (e.g. "- Made a change (#0 by @)")
- [x] Linked any existing issues or proposals that this pull request should close
- [ ] Updated or added relevant documentation
- [ ] Added a test for the contribution (if applicable)

PR #297, opened 2022-08-13
Revert shim mode

JordanMartinez:
**Description of the change**

Reverts #295. See https://github.com/purescript/trypurescript/pull/294#issuecomment-1213615468 and I'm still waiting for Andy to confirm.

---

**Checklist:**

### DIFF
--- a/client/public/frame.html
+++ b/client/public/frame.html
@@ -5,11 +5,6 @@
   <meta content="text/html;charset=utf-8" http-equiv="Content-Type">
   <meta content="utf-8" http-equiv="encoding">
   <meta name="viewport" content="width=device-width, initial-scale=1.0">
-  <script>
-    window.esmsInitOptions = {
-      shimMode: true,
-    };
-  </script>
   <!--
     JSPM Generator Import Map
     Edit URL: https://generator.jspm.io/#ZcxBDkAwEEDRWQinMQwhlr2EA9BOZKQqqamtq5OIle1L/s8zgOIag4p6djDLUkpQXjgawh47AsdWtsnjehiqsUWCyJPV0u2boQEbrH9QWS8c9O8Hx5Pj61+ckjgzPOPmBszFaRmMAA


andys8:
Issues are very likely depending on browser *and* version. So, would be good to add that information. I'll have a look tomorrow and compare browsers in combination with `es-module-shims` settings.

JordanMartinez:
Issue is on Firefox 103.0.1 on a Linux OS. I haven't checked on Chrome.

andys8:
# Poor man's browserstack testing 😉

*Work in Progress*

## Browsers

- Top Left: Chromium 104
- Bottom Left: Qutebrowser 2.5.2 based on Chromium 87
- Top Right: Firefox 103
- Bottom Right: Pale Moon 32 based on Firefox

## Example 1

```purescript
module HelloReactHooks.Main where

import Prelude

import Data.Maybe (Maybe(..))
import Effect (Effect)
import Effect.Exception (throw)
import React.Basic.DOM as R
import React.Basic.DOM.Client (createRoot, renderRoot)
import React.Basic.Hooks (Component, component)
import Web.DOM.NonElementParentNode (getElementById)
import Web.HTML (window)
import Web.HTML.HTMLDocument (toNonElementParentNode)
import Web.HTML.Window (document)

main :: Effect Unit
main = do
  doc <- document =<< window
  root <- getElementById "root" $ toNonElementParentNode doc
  case root of
    Nothing -> throw "Could not find root."
    Just container -> do
      reactRoot <- createRoot container
      app <- mkApp
      renderRoot reactRoot (app {})

mkApp :: Component {}
mkApp = do
  component "App" \_ -> React.do
    pure (R.text "Hello!")
```

## Example 2

```purescript
module Main where

import Prelude

import Data.Foldable (fold)
import Effect (Effect)
import TryPureScript (h1, h2, p, text, list, indent, link, render, code)

main :: Effect Unit
main = render $ h1 $ text "a"
```

## Example 1, `shimMode: false`

image

## Example 2, `shimMode: false`

image

## Example 1, `shimMode: true`

image

## Example 2, `shimMode: true`

image

## Example 1, `import uuid from `uuid``

image

## Example 2, `import uuid from `uuid``

image




`

andys8:
# Solution (?)

## Updating `es-module-shims` to `1.5.12`

No shim mode, no settings, no import

```html
<script async src="https://ga.jspm.io/npm:es-module-shims@1.5.12/dist/es-module-shims.js" crossorigin="anonymous"></script>
```

## Example 1

image


## Example 2

image

## Summary

Updating es-module-shims to `1.5.12` seems to be a valid solution for the issues I could reproduce.

## Additional safety?

That being said, I'm a bit sceptible that I couldn't reproduce the double rendering issue. In case it's possible to reproduce it somehow (after the update), I would recommend adding an additional import script, in order to force the polyfills to kick in (what is described [here](https://github.com/guybedford/es-module-shims#usage): "The first import being a bare specifier in the pattern above is important to ensure this.")

```
  <script type="module">
    import uuid from 'uuid';
  </script>
```

The README example imports react and logs it. To me this looks weird, and I'd prefer to avoid it because the logs including errors are confusing for the users. But if necessary, I would just do it and add a comment to the implementation, instead of further spending time on this issue.

andys8:
# Comparing `es-module-shims` versions

No import, no shimMode, Example 1

## 1.5.1

image


## 1.5.9

image


## 1.5.12 :heavy_check_mark: 

image

## 1.5.13

image

## 1.5.14

image


## Summary

This is terrible. "Solution" is to pin to 1.5.12 and add a comment to only update if tested.

natefaubion:
Is it worth opening a bug ticket and asking for a recommendation?

andys8:
@natefaubion Opened an issue https://github.com/guybedford/es-module-shims/issues/324 (not the best description, but I'm  running out of time) 😉

andys8:
I suggest to go with 1.5.12 for now: https://github.com/purescript/trypurescript/pull/298

See https://github.com/guybedford/es-module-shims/issues/324#issuecomment-1214452714

andys8:
## Update

I'm pretty sure the the issues with the versions 1.5.13 - 1.5.16 are related to this code.

https://github.com/purescript/trypurescript/blob/3cdcc8ff97a9089efc522dfed7d55637f3c3ce97/client/public/js/frame.js#L13-L14

The event listeners are initialized with `once: true`. I think what we see with the browsers with issues (e.g. latest firefox) is that the import machinery of es-module kicks in, and triggers re-initialization, after polyfills are applied. Because of this flag it's not executed a second time. Removing these will execute the inner function twice for polyfilled browsers, but it'll show up and work for the user (tested with 1.5.16), although double execution and import errors showing up multiple times in the log.

There is also a [`noLoadEventRetriggers`](https://github.com/guybedford/es-module-shims#no-load-event-retriggers) setting in `es-module-shims`, but I couldn't see any effect or difference for this project, when I tried it.

natefaubion:
Can we serve a noop import map module and always import that first, if it fixes the double load issue?

andys8:
> Can we serve a noop import map module and always import that first, if it fixes the double load issue?

Tried importing a library of the ones in the importmap, didn't make a difference. I assume this is what you mean with "noop import map"?

* Also tried `async` or not on several scripts
* Different order of scripts in `frame.html`
* Different options of `es-module-shims`
* Changing `frame.js` to being a module
* Changing `frame.js` to import one of the libraries
* Manually making sure event listeners either get removed or are only set once

JordanMartinez:
Superceded by #298

